PR #168, opened 2023-07-24
removing statefulset specific code, reworking code for cnpg

nhudson:
More code deletion in this PR
* Remove `StatefulSet` specific code.  We will no longer reconcile or "know" about the `StatefulSet` that is created.  Effectively we will just start ignoring it.
* Remove the old coredb `Service` that was associated with connecting to the old CoreDB instance.
* Change default image logic a bit.  We will now take the default image set in `cdb.spec.image` or if that is empty we will look for a stack and set it correctly or use the default image.
* Fix tests

Before merging this, we will need to do a few other things.
* Update the `image` to use the new CNPG specific defaults in the [cp-webserver](https://github.com/tembo-io/control-plane/tree/3d1f267f57d3d888b23bbafd6c20cde8be9a3d61/cp-webserver/src/stacks)


nhudson:
Some notes to follow for after this PR is merged

To upgrade coredb-operator we will need to manually adjust the following.

1: All manifests will need to be updated to use the correct cnpg image. (tembo-pg-cnpg or ml-cnpg) only.

    * There could be errors when reconciling the image back to the coredb-pg-slim image.

2: Manually delete the `CronJobs` in each namespace.  The CJ backups are no longer needed

3: Manually remove the `Service` in each namespace.  It will be the only one without a suffix.

4: Manually remove the `StatefulSet` in each namespace

5: Manually remove the PVC's in each namespace that were associated with each `StatefulSet`.  There should be 3 in each namespace.

nhudson:
Conductor tests will most likely fail until we can get this PR merged.  This is due to the fact that Conductor tests are ran from `main` code base of the operator and not this PR's code base.

tembo-operator/src/extensions.rs
```diff
@@ -192,14 +173,11 @@ pub async fn install_extensions(
...
-                let cnpg_exec = cdb.exec(pod_name_cnpg.clone().unwrap().clone(), client.clone(), &cmd);
-                let _ = coredb_exec.await;
-                cnpg_exec.await
-            }
-            false => coredb_exec.await,
+        let result = if let Some(pod_name) = pod_name_cnpg.clone() {
+            let cnpg_exec = cdb.exec(pod_name, client.clone(), &cmd);
+            cnpg_exec.await
+        } else {
+            continue;
```
sjmiller609, line 180: instead of if let Some else on pod_name_cnpg, can you edit above so the assignment
```
let pod_name_cnpg = ...
```

is type String, and when primary_pod_cnpg has returned an Err, just return that?

So, I think that is done something like this

```
    let pod_name_cnpg: String = cdb
                .primary_pod_cnpg(client.clone())
                .await?
                .metadata
                .name
                .expect("Pod should always have a name");
```

The `await?` part should requeue with the Action returned by the primary_pod_cnpg (for example pod is retarting)

sjmiller609 (approved):
One nit

tembo-operator/src/extensions.rs
nhudson: We are already doing something like this right above it, [here](https://github.com/tembo-io/tembo-stacks/blob/088b92c563c724879afb66339e2add3681bf4020/tembo-operator/src/extensions.rs#L145-L156)

sjmiller609: Yeah, I mean replace those lines so it's type String instead of Option<String>, since now we would want to exit and requeue if we do not find the primary pod of CNPG

nhudson: Hopefully this last update will be what you're looking for.

tembo-operator/src/extensions.rs
```diff
@@ -143,41 +142,24 @@ pub async fn install_extensions(
...
-        }
-        false => None,
+    let pod_name_cnpg: String = match cnpg_enabled {
+        true => cdb
+            .primary_pod_cnpg(client.clone())
+            .await?
+            .metadata
+            .name
+            .expect("Pod should always have a name"),
+        false => "".to_string(),
```
nhudson, line 152: @sjmiller609 changed this to just be a `String`

tembo-operator/src/extensions.rs
```diff
@@ -143,41 +142,24 @@ pub async fn install_extensions(
...
 
     let mut errors: Vec<Error> = Vec::new();
     let num_to_install = extensions.len();
     for ext in extensions.iter() {
-        let version = ext.locations[0].version.clone().unwrap();
+        let version = ext.locations[0].version.clone().unwrap_or_else(|| {
+            let err = Error::InvalidErr("Missing version for extension".to_string());
+            error!("{}", err);
+            err.to_string()
+        });
```
nhudson, line 162: Adding better error logic here.  Before if the `unwrap()` errored it would have caused a crash.  Now at least we can handle the error and move on.

tembo-operator/src/extensions.rs
```diff
@@ -271,17 +253,18 @@ pub async fn toggle_extensions(
 
                 match result {
                     Ok(_) => {}
-                    Err(_) => {
+                    Err(e) => {
```
nhudson, line 256: Match block is too generic, lets make sure to handle the error correctly.

tembo-operator/src/extensions.rs
sjmiller609: I think if CNPG is not enabled on line 143, just requeue, and do not match on cnpg_enabled here, setting the pod name to empty string does not seem right

nhudson: Yeah agreed, I missed  that.  Should be fixed now.

sjmiller609 (approved):

PR #179, opened 2023-07-27
Fix dashboards so that Grafana will pick them up

nhudson:


nhudson:
[TEM-1363](https://linear.app/tembo/issue/TEM-1363/fix-grafana-dashboards)

PR #181, opened 2023-07-31
Update rust version conductor

sjmiller609:
```
error: failed to compile `conductor v0.1.0 (/build)`, intermediate artifacts can be found at `/build/target`

Caused by:
  package `time v0.3.24` cannot be built because it requires rustc 1.67.0 or newer, while the currently active rustc version is 1.66.1
  Either upgrade to rustc 1.67.0 or newer, or use
  cargo update -p time@0.3.24 --precise ver
  where `ver` is the latest version of `time` supporting rustc 1.66.1
The command '/bin/sh -c cargo install --path .' returned a non-zero code: 101
```

### DIFF
--- a/conductor/Dockerfile
+++ b/conductor/Dockerfile
@@ -6,10 +6,11 @@ RUN apt-get update && \
 
 COPY tembo-operator ./tembo-operator
 WORKDIR /build
-COPY Cargo.toml Cargo.lock ./
-COPY src ./src
-RUN cargo build --release && \
-    cargo clean -p conductor
+
+COPY Cargo.toml .
+COPY Cargo.lock .
+COPY ./src ./src
+
 RUN cargo install --path .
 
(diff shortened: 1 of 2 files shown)


nhudson (approved):

conductor/Dockerfile
```diff
@@ -1,10 +1,16 @@
...
+
+RUN apt-get update && \
+    apt-get install -y pkg-config libssl-dev && apt-get clean && \
+    rm -rf /var/lib/apt/lists/* /var/cache/apt/archives/*
+
 COPY tembo-operator ./tembo-operator
 WORKDIR /build
-COPY Cargo.toml Cargo.lock ./
-COPY src ./src
-RUN cargo build --release && \
```
sjmiller609, line 6: with build --release and install, it's just running the build twice

PR #184, opened 2023-08-01
Add support for enabling traces

nhudson:
Adding support for enabling the tracing feature.

* Our controller already has a compile time flag to enable tracing, you must set `OPENTELEMETRY_ENDPOINT_URL` as an env var.
* Updating `Dockerfile` to use latest stable version of Rust and Alpine
* Update `telemetry.rs` code and crates for better telemetry and trace pattern usage.
* Update k8s-openapi and kube-rs crates
* Update code to work with updates crates.


nhudson:
Added `OPENTELEMETRY_ENDPOINT_URL` environment variable to the Helm chart deployment https://github.com/tembo-io/app-deploy-dev/commit/df712eaef2dfc6293ac804daf5085cf619be70fd

This will need to be added for staging and production deployments as well before we merge.

sjmiller609 (approved):

PR #185, opened 2023-08-02
Fix build issues with latest version of tembo-operator

nhudson:
With the tracing changes to `tembo-operator` we had to bump the `kube` and `k8s-openapi` crate.  Since both `conductor` and `tembo-pod-init` use the `controller` crate we also need to bump those crates in each of these projects to be compatable.


nhudson:
Keep in mind that conductor functional tests are still failing.  This should be fixed soon.

sjmiller609 (approved):

PR #195, opened 2023-08-08
migrate to use tembo-telemetry

nhudson:
Migrate project to use [tembo-telemetry](https://github.com/tembo-io/tembo-telemetry)


nhudson:
I believe we will need to add this feature/option to the tembo-telemetry crate before we can do this https://github.com/tembo-io/tembo-telemetry/issues/2

PR #201, opened 2023-08-09
Deleting unused code, fix config map updates

sjmiller609:


tembo-operator/src/apis/postgres_parameters.rs
```diff
@@ -371,34 +361,6 @@ impl<'de> Deserialize<'de> for PgConfig {
     }
 }
 
-pub async fn reconcile_pg_parameters_configmap(
```
sjmiller609, line 374: this was for before CNPG migration, now this is not used

tembo-operator/src/configmap.rs
```diff
@@ -1,59 +1,14 @@
...
-    api::{Api, ObjectMeta, Patch, PatchParams, PostParams},
+    api::{Api, ObjectMeta, Patch, PatchParams},
     Client,
 };
 use std::collections::BTreeMap;
 
-use tracing::{debug, error, info, warn};
+use tracing::{debug, error};
 
-pub async fn create_configmap_ifnotexist(
```
sjmiller609, line 11: We should not create empty configmaps

tembo-operator/src/postgres_exporter.rs
```diff
@@ -13,7 +7,7 @@ use tracing::{debug, error};
 
 pub const QUERIES_YAML: &str = "queries.yaml";
 pub const EXPORTER_VOLUME: &str = "postgres-exporter";
-pub const EXPORTER_CONFIGMAP: &str = "postgres-exporter";
+pub const EXPORTER_CONFIGMAP_PREFIX: &str = "metrics-";
```
sjmiller609, line 10: configmaps should have unique names because there could be multiple in the same namespace

tembo-operator/src/controller.rs
```diff
@@ -350,6 +259,7 @@ impl CoreDB {
 
         apply_status_force(&coredbs, &name, patch_status).await?;
 
+        info!("Fully reconciled {}", self.name_any());
```
sjmiller609, line 262: I'm noticing on local that even though this log shows up, it immediately does another reconcile

nhudson (approved):

PR #203, opened 2023-08-09
fix logic where we add metrics volume

nhudson:
The previous logic will always add a volume to the deployment mainly because `cdb.spec.metrics` is never empty by default, or at least shouldn't be.  New logic should handle the case correctly, where we only want to mount the `ConfgMap` if there are configured queries.

```
spec:
  metrics:
    enabled: true
    image: quay.io/prometheuscommunity/postgres-exporter:v0.12.0
```

### DIFF
--- a/tembo-operator/src/deployment_postgres_exporter.rs
+++ b/tembo-operator/src/deployment_postgres_exporter.rs
@@ -22,7 +22,6 @@ use kube::{
 };
 use std::{collections::BTreeMap, sync::Arc};
 
-
 const PROM_CFG_DIR: &str = "/prometheus";
 
 pub async fn reconcile_prometheus_exporter_deployment(cdb: &CoreDB, ctx: Arc<Context>) -> Result<(), Error> {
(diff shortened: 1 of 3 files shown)


tembo-operator/src/deployment_postgres_exporter.rs
```diff
@@ -140,11 +139,8 @@ pub async fn reconcile_prometheus_exporter_deployment(cdb: &CoreDB, ctx: Arc<Con
     };
 
     // Generate Volumes for the PodSpec
-    let exporter_volumes = match cdb.spec.metrics {
-        None => {
-            vec![]
-        }
-        Some(_) => {
+    let exporter_volumes = if let Some(metrics) = &cdb.spec.metrics {
```
sjmiller609, line 142: Nice.

Also maybe could do the same where the configmap is created?

sjmiller609 (approved):

tembo-operator/src/deployment_postgres_exporter.rs
nhudson: We kind of do this already.  looks like we always call `reconcile_prom_configmap` but it will do nothing if `m.queries` do not exist.

```
    match cdb.spec.metrics.clone().and_then(|m| m.queries) {
        Some(queries) => {
            let qdata = serde_yaml::to_string(&queries).unwrap();
            let d: BTreeMap<String, String> = BTreeMap::from([(QUERIES_YAML.to_string(), qdata)]);
            apply_configmap(
                client.clone(),
                ns,
                &format!("{}{}", EXPORTER_CONFIGMAP_PREFIX, coredb_name),
                d,
            )
            .await?
        }
        None => {
            debug!("No queries specified in CoreDB spec");
        }
    }
    Ok(())
```

nhudson: I think we could improve this to just never call the function.

sjmiller609: Ok you're right, seems like it's good already then

PR #204, opened 2023-08-09
Fix issue with status flapping

sjmiller609:
Status was flipping back and forth from having error_message as null or not included.


tembo-operator/src/controller.rs
```diff
@@ -263,7 +263,7 @@ impl CoreDB {
             "status": new_status
         });
 
-        apply_status_force(&coredbs, &name, patch_status).await?;
+        patch_cdb_status_merge(&coredbs, &name, patch_status).await?;
```
sjmiller609, line 266: Since we set every option of CoreDBStatus, this has no difference in effect

nhudson (approved):

PR #206, opened 2023-08-11
Continuous status reporting

sjmiller609:
Watch for changes in resources, and report back to control plane any change

### DIFF
--- a/conductor/src/lib.rs
+++ b/conductor/src/lib.rs
@@ -51,15 +51,18 @@ pub async fn generate_spec(
     })
 }
 pub fn get_data_plane_id_from_coredb(coredb: &CoreDB) -> Result<String, Box<ConductorError>> {
-    let annotations = coredb
-        .metadata
-        .annotations
-        .as_ref()
-        .ok_or(ConductorError::EventIDFormat)?;
-    let data_plane_id = annotations
-        .get("tembo.io/data_plane_id")
-        .ok_or(ConductorError::EventIDFormat)?
-        .to_string();
+    let annotations = match coredb.metadata.annotations.as_ref() {
+        None => {
+            return Err(Box::new(ConductorError::EventIDFormat));
+        }
+        Some(annotations) => annotations,
+    };
+    let data_plane_id = match annotations.get("tembo.io/data_plane_id") {
+        Some(data_plane_id) => data_plane_id.to_string(),
+        None => {
+            return Err(Box::new(ConductorError::EventIDFormat));
+        }
+    };
     Ok(data_plane_id)
 }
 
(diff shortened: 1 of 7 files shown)


conductor/src/lib.rs
```diff
@@ -24,16 +24,82 @@ use serde_json::{from_str, to_string, Value};
...
         "kind": "CoreDB",
         "metadata": {
             "name": namespace,
+            "annotations": {
+                "tembo.io/org_id": org_id,
+                "tembo.io/instance_id": instance_id,
+                "tembo.io/workspace_id": workspace_id,
+                "tembo.io/entity_name": entity_name,
+                "tembo.io/data_plane_id": data_plane_id,
+            }
```
sjmiller609, line 47: storing some new metadata in annotations - I think this seems nicer in comparison to parsing the org out from the namespace formatting, for example.

conductor/src/routes/health.rs
```diff
@@ -0,0 +1,22 @@
...
+        }
+    };
+    for thread in background_threads.iter() {
+        if thread.is_finished() {
+            return HttpResponse::InternalServerError()
+                .body("One or more background tasks are not running.");
+        }
+    }
+    HttpResponse::Ok().json("ok")
+}
```
sjmiller609, line 22: I used this same pattern in control plane to fail liveness checks if a background thread that's supposed to be continuously running fails.

conductor/src/lib.rs
```diff
@@ -24,16 +25,82 @@ use serde_json::{from_str, to_string, Value};
...
+}
+pub fn get_data_plane_id_from_coredb(coredb: &CoreDB) -> Result<String, ConductorError> {
+    let annotations = coredb
+        .metadata
+        .annotations
+        .as_ref()
+        .ok_or(ConductorError::EventIDParsing)?;
+    let data_plane_id = annotations
+        .get("tembo.io/data_plane_id")
+        .ok_or(ConductorError::EventIDParsing)?
```
sjmiller609, line 61:
```suggestion
        .ok_or(ConductorError::EventIDFormat)?;
    let data_plane_id = annotations
        .get("tembo.io/data_plane_id")
        .ok_or(ConductorError::EventIDFormat)?
```

conductor/src/status_reporter.rs
```diff
@@ -0,0 +1,123 @@
...
+
+    // Get a kubernetes watcher on all changes in coredb resources
+    let client = Client::try_default().await?;
+    let coredb_api: Api<CoreDB> = Api::all(client.clone());
+
+    watcher(coredb_api, watcher::Config::default())
+        .applied_objects()
+        .try_for_each(move |coredb| {
+            let client = client.clone();
+            let queue = queue.clone();
```
sjmiller609, line 31: I need to make sure the queue being cloned doesn't result in another connection - otherwise this will cause large bursts of pg connections

nhudson (approved):

PR #210, opened 2023-08-17
Add PrometheusRules

nhudson:
Add `PrometheusRule` for CNPG based instances.

### DIFF
--- a/charts/tembo-operator/templates/cnpg-prometheusrules.yaml
+++ b/charts/tembo-operator/templates/cnpg-prometheusrules.yaml
@@ -2,7 +2,7 @@
 apiVersion: monitoring.coreos.com/v1
 kind: PrometheusRule
 metadata:
-  name: {{ .Release.name }}-cnpg-alerts
+  name: {{ .Release.Name }}-cnpg-alerts
   namespace: {{ .Release.Namespace }}
 spec:
   groups:
--- a/charts/tembo-operator/templates/cnpg-prometheusrules.yaml
+++ b/charts/tembo-operator/templates/cnpg-prometheusrules.yaml
@@ -1,0 +1,73 @@
+{{- if .Values.prometheusRule.enabled -}}
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: {{ .Release.name }}-cnpg-alerts
+  namespace: {{ .Release.Namespace }}
+spec:
+  groups:
+  - name: cnpg.rules
+    rules:
+    - alert: LongRunningTransaction
+      annotations:
+        description: Pod {{`{{`}} $labels.pod {{`}}`}} is taking more than 5 minutes (300 seconds) for a query.
+        summary: A query is taking longer than 5 minutes.
+      expr: |-
+        cnpg_backends_max_tx_duration_seconds > 300
+      for: 1m
+      labels:
+        severity: warning
+    - alert: BackendsWaiting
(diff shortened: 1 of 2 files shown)


charts/tembo-operator/templates/cnpg-prometheusrules.yaml
```diff
@@ -0,0 +1,73 @@
...
+        severity: warning
+    - alert: LastFailedArchiveTime 
+      annotations:
+        description: Archiving failed for {{`{{`}} $labels.pod {{`}}`}}
+        summary: Checks the last time archiving failed. Will be < 0 when it has not failed.
+      expr: |-
+        (cnpg_pg_stat_archiver_last_failed_time - cnpg_pg_stat_archiver_last_archived_time) > 1
+      for: 1m
+      labels:
+        severity: warning
```
sjmiller609, line 55: Can you check off this box and link to this code change https://www.notion.so/tembo-io/SOC2-OPs-Engineering-Async-Doc-8ae85f83ede040f49252adf73a2a4e39?pvs=4#692631e1a491412f825658a6096d50f5

Also, is it possible to also alert on the inverse, like alert when number of backups for any given DB is less than 1 (i.e. catch the situation if backups not running or being attempted to run for some reason).

charts/tembo-operator/templates/cnpg-prometheusrules.yaml
```diff
@@ -0,0 +1,73 @@
...
+    - alert: PostgreSQLFailoverDetected
+      expr: changes(cnpg_pg_replication_in_recovery[2m])
+        > 0
+      labels:
+        severity: critical
+      annotations:
+        summary: "PostgreSQL instance failover detected"
+        description: "The PostgreSQL instance in {{`{{`}} $labels.namespace {{`}}`}}/{{`{{`}} $labels.pod {{`}}`}} transitioned its recovery state, indicating a potential failover."
```
sjmiller609, line 72: I'm not sure if this would be triggered for intentional "switch overs", for example during node rotation, instance restart, and so on - is that a different process?

sjmiller609 (approved):
excellent

ianstanton (approved):

charts/tembo-operator/templates/cnpg-prometheusrules.yaml
sjmiller609: I mean consider if we accidentally disabled backups somehow, what would fire?

charts/tembo-operator/templates/cnpg-prometheusrules.yaml
nhudson: This technically should fire for every switch over.  My thought was to enable it and make sure it works, then refine it.  Though I am not sure how or what else to refine it with.  There isn't a really good way to detect when a failover happens either from what I was testing yesterday.

charts/tembo-operator/templates/cnpg-prometheusrules.yaml
nhudson: From what I can tell, they don't provide a count of the number of backups.  Their metrics expose only three backup metrics: `cnpg_collector_last_available_backup_timestamp`, `cnpg_collector_last_failed_backup_timestamp` and `cnpg_collector_first_recoverability_point`.

We could look at last available backup timestamp, then maybe run a calculation to see if it's over a specific time, like 2 hours or something.  It should take a backup once an hour.

PR #213, opened 2023-08-17
Enable HA/Replication Slots

nhudson:
This PR is meant to add the ability to enable HA/Replication Slots configuration into the Tembo Cloud stack.

* The operator will now automatically enable HA when 2 or more replicas are set in the `CoreDBSpec`.
* We now respect the value of `replicas` and it will set the replica/instance count from `spec.replicas`

Fixes: [TEM-1384](https://linear.app/tembo/issue/TEM-1384)

### DIFF
--- a/tembo-operator/src/cloudnativepg/cnpg.rs
+++ b/tembo-operator/src/cloudnativepg/cnpg.rs
@@ -10,8 +10,9 @@ use crate::{
             ClusterExternalClusters, ClusterExternalClustersPassword, ClusterLogLevel, ClusterMonitoring,
             ClusterMonitoringCustomQueriesConfigMap, ClusterNodeMaintenanceWindow, ClusterPostgresql,
             ClusterPostgresqlSyncReplicaElectionConstraint, ClusterPrimaryUpdateMethod,
-            ClusterPrimaryUpdateStrategy, ClusterResources, ClusterServiceAccountTemplate,
-            ClusterServiceAccountTemplateMetadata, ClusterSpec, ClusterStorage, ClusterSuperuserSecret,
+            ClusterPrimaryUpdateStrategy, ClusterReplicationSlots, ClusterReplicationSlotsHighAvailability,
+            ClusterResources, ClusterServiceAccountTemplate, ClusterServiceAccountTemplateMetadata,
+            ClusterSpec, ClusterStorage, ClusterSuperuserSecret,
         },
         scheduledbackups::{
             ScheduledBackup, ScheduledBackupBackupOwnerReference, ScheduledBackupCluster, ScheduledBackupSpec,
@@ -235,20 +236,32 @@ fn cnpg_cluster_storage(cdb: &CoreDB) -> Option<ClusterStorage> {
     })
 }
 
+// Check replica count to enable HA
+fn cnpg_high_availability(cdb: &CoreDB) -> Option<ClusterReplicationSlots> {
+    if cdb.spec.replicas > 1 {
+        Some(ClusterReplicationSlots {
+            high_availability: Some(ClusterReplicationSlotsHighAvailability {
+                enabled: Some(true),
+                ..ClusterReplicationSlotsHighAvailability::default()
+            }),
+            update_interval: Some(30),
+        })
+    } else {
+        None
+    }
+}
+
 pub fn cnpg_cluster_from_cdb(cdb: &CoreDB) -> Cluster {
     let cfg = Config::default();
     let name = cdb.name_any();
     let namespace = cdb.namespace().unwrap();
     let owner_reference = cdb.controller_owner_ref(&()).unwrap();
-
     let mut annotations = BTreeMap::new();
     annotations.insert("tembo-pod-init.tembo.io/inject".to_string(), "true".to_string());
-
     let (bootstrap, external_clusters, superuser_secret) = cnpg_cluster_bootstrap_from_cdb(cdb);
-
     let (backup, service_account_template) = cnpg_backup_configuration(cdb, &cfg);
-
     let storage = cnpg_cluster_storage(cdb);
+    let replication = cnpg_high_availability(cdb);
 
     let PostgresConfig {
         postgres_parameters,
@@ -297,7 +310,7 @@ pub fn cnpg_cluster_from_cdb(cdb: &CoreDB) -> Cluster {
             enable_superuser_access: Some(true),
             failover_delay: Some(0),
             image_name: Some(image),
-            instances: 1,
+            instances: cdb.spec.replicas as i64,
             log_level: Some(ClusterLogLevel::Info),
             max_sync_replicas: Some(0),
             min_sync_replicas: Some(0),
@@ -325,6 +338,7 @@ pub fn cnpg_cluster_from_cdb(cdb: &CoreDB) -> Cluster {
             }),
             primary_update_method: Some(ClusterPrimaryUpdateMethod::Restart),
             primary_update_strategy: Some(ClusterPrimaryUpdateStrategy::Unsupervised),
+            replication_slots: replication,
             resources: Some(ClusterResources {
                 claims: None,
                 limits: cdb.spec.resources.clone().limits,
--- a/tembo-operator/src/cloudnativepg/cnpg.rs
+++ b/tembo-operator/src/cloudnativepg/cnpg.rs
@@ -247,7 +247,13 @@ fn cnpg_high_availability(cdb: &CoreDB) -> Option<ClusterReplicationSlots> {
             update_interval: Some(30),
         })
     } else {
-        None
+        Some(ClusterReplicationSlots {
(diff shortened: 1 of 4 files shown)


charts/tembo-operator/templates/crd.yaml
```diff
@@ -82,6 +82,14 @@ spec:
...
                   type: object
                 type: array
+              high_availability:
+                nullable: true
+                properties:
+                  enable_replication_slots:
+                    default: false
+                    nullable: true
+                    type: boolean
+                type: object
```
sjmiller609, line 92: Let's just pick what we think is best under the hood when replicas > 1 and not allow configuration of it in CoreDB spec, so we can remove this section

charts/tembo-operator/templates/crd.yaml
```diff
@@ -128,7 +136,7 @@ spec:
                 type: string
               replicas:
                 default: 1
-                format: int32
+                format: int64
```
sjmiller609, line 139: Let's just type cast this in the operator instead of changing the crd, since the type is included in the CRD spec and I don't know exactly what happens to existing resources if that's changed - but could be convinced otherwise

tembo-operator/src/cloudnativepg/cnpg.rs
```diff
@@ -235,20 +236,37 @@ fn cnpg_cluster_storage(cdb: &CoreDB) -> Option<ClusterStorage> {
...
 }
 
+// Check for HA and number of replicas.  Retrun (None, 1) by default
+fn cnpg_high_availability(cdb: &CoreDB) -> Option<ClusterReplicationSlots> {
+    match &cdb.spec.high_availability {
+        Some(ha) => {
+            match ha.enable_replication_slots {
+                Some(true) => Some(ClusterReplicationSlots {
+                    high_availability: Some(ClusterReplicationSlotsHighAvailability {
+                        enabled: Some(true),
```
sjmiller609, line 246: https://github.com/cloudnative-pg/cloudnative-pg/pull/2601

tembo-operator/tests/integration_tests.rs
```diff
@@ -1437,4 +1437,98 @@ mod test {
         // Delete namespace
         let _ = delete_namespace(client.clone(), &namespace).await;
     }
+
+    #[tokio::test]
+    #[ignore]
+    async fn functional_test_ha_cnpg() {
```
sjmiller609, line 1443: Can you please expand this test or add a different test to initially launch with replicas 1, assert looks good, then change replicas to 2 and do the rest

tembo-operator/src/cloudnativepg/cnpg.rs
```diff
@@ -297,7 +317,7 @@ pub fn cnpg_cluster_from_cdb(cdb: &CoreDB) -> Cluster {
             enable_superuser_access: Some(true),
             failover_delay: Some(0),
             image_name: Some(image),
-            instances: 1,
+            instances: cdb.spec.replicas,
             log_level: Some(ClusterLogLevel::Info),
             max_sync_replicas: Some(0),
             min_sync_replicas: Some(0),
```
sjmiller609, line 323: 👍

tembo-operator/src/cloudnativepg/cnpg.rs
sjmiller609: When I check the CRD in our cluster, the default is false.
```
              replicationSlots:
                description: Replication slots management configuration
                properties:
                  highAvailability:
                    description: Replication slots for high availability configuration
                    properties:
                      enabled:
                        default: false
```

The helm chart is not updated to match the docs https://github.com/cloudnative-pg/charts/blob/39c99bc63cb2f97e2d2895484da2bbec0ee6c71f/charts/cloudnative-pg/templates/crds/crds.yaml#L3154

I am concerned because there is a note that this can only be enabled at creation time of the cluster

sjmiller609: I'm not sure what should be done about that yet, but we should check in on this case where it's set to false, then we switch it to true on version 1.20 (which we already have) - maybe the docs are out of date, I'm not sure what would happen.

The ideal fix would be CNPG helm chart and docs updated to the correct status, then we don't specify this in our clusters, since it's just using the default configuration.

sjmiller609 (changes requested):

tembo-operator/src/cloudnativepg/cnpg.rs
```diff
@@ -286,7 +304,9 @@ pub fn cnpg_cluster_from_cdb(cdb: &CoreDB) -> Cluster {
         spec: ClusterSpec {
             affinity: Some(ClusterAffinity {
                 pod_anti_affinity_type: Some("preferred".to_string()),
-                topology_key: Some("topology.kubernetes.io/zone".to_string()),
+                // topology_key will be different across clouds.  For now just set it to what it
+                // needs to be for AWS EKS
+                topology_key: Some("failure-domain.beta.kubernetes.io/zone".to_string()),
```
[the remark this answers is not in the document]
nhudson: Weird I didn't yesterday when I looked. If so then we can change it back

tembo-operator/src/cloudnativepg/cnpg.rs
nhudson: This is an async replication configuration. If we ever want to enable synchronous replication, then we will need to do some calculations around how many replicas are set.

tembo-operator/src/cloudnativepg/cnpg.rs
nhudson: Not really sure what the issue is here. The default is false. I do believe from reading the docs if you set replicas to 2 it will enable this automatically unless you specifically tell it not to. I didn't test that scenario since we would always want this enabled if the cluster is in HA mode .

tembo-operator/src/cloudnativepg/cnpg.rs
sjmiller609: Ok sounds good. I prefer async and think it makes sense for what HA is supposed to do (increase SLA), maybe some customers in the future want sync for their use case (extreme durability requirement in comparison to performance and availability)

tembo-operator/src/cloudnativepg/cnpg.rs
sjmiller609: I mean since it's currently default false, and we already have clusters deployed, we are trying to switch them to true, but the feature has a note that it's only allowed to be set at creation

https://cloudnative-pg.io/documentation/1.20/api_reference/#ReplicationSlotsHAConfiguration

> This can only be set at creation time.

nhudson: I would say that is wrong then since I am able to set it and update it to HA from a single instance that was already created. 

Maybe it's true for CNPG operator version <1.20 since this is a new feature in 1.20?

sjmiller609: Ok that is what I’m hoping is the case, you should mention that in the cnpg PR

tembo-operator/tests/integration_tests.rs
nhudson: Added a test to launch a single instance and then update it to run with 2 replicas.

charts/tembo-operator/templates/crd.yaml
nhudson: This has been reverted.

tembo-operator/src/cloudnativepg/cnpg.rs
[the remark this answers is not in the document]
nhudson: reverted

sjmiller609 (approved):
Looks good!

Can you please QA this situation in dev:

Launch new instance, enable HA, then disable HA (this will move the replication slots management from false -> true -> false). Check if replication slots are managed (removed), i.e. checking they are not just filling up the replication slot. Since if the replication slots are not removed, then they will fill up and cause disk to fill up

nhudson:
> Looks good! I think OK to merge if tests pass, but needs some special attention in dev cluster QA.
> 
> Can you please QA this situation in dev:
> 
> Launch new instance, enable HA, check replication slots are created. Then disable HA (this will move the replication slots management from false -> true -> false). Check if replication slots are managed (removed), i.e. checking they are not just filling up the replication slot. Since if the replication slots are not removed, then they will fill up and cause disk to fill up

I am looking into the failing test now.  

For your scenario @sjmiller609, if you disable HA (set replicas to 1), it removes the replication slots entry from the table
```
postgres=# SELECT * FROM pg_replication_slots;
 slot_name | plugin | slot_type | datoid | database | temporary | active | active_pid | xmin | catalog_xmin | restart_lsn | confirmed_flush_lsn | wal_status | safe_wal_size | two_phase 
-----------+--------+-----------+--------+----------+-----------+--------+------------+------+--------------+-------------+---------------------+------------+---------------+-----------
(0 rows)
```

PR #216, opened 2023-08-21
Add connection string info for app user

nhudson:
This PR changes how we send back connection string information to the control-plane.  This will include the connection string info for both the `postgres` and `app` users. 

Some things to note for this change:  This will most likely break Mahout and rendering of the connection info.  I am not sure what would need to be changed to get that working with this change yet.

* Changes the response back to the control-plane from a single `ConnectionInfo` into a tuple of `ConnectionInfo`
* Added `Box` to the `AwsErr` trait to keep clippy from warning about large variants
* Reworked how we extract the secret data to better handle multiple secrets if we wish to add more in the future.
* Added helper functions to better handle data type conversion from the Kubernetes Secret to create `ConnectionInfo`

fixes: [TEM-1389](https://linear.app/tembo/issue/TEM-1389/add-connection-strings-for-app-user)

### DIFF
--- a/conductor/src/lib.rs
+++ b/conductor/src/lib.rs
@@ -21,7 +21,7 @@ use chrono::{SecondsFormat, Utc};
 use kube::{Api, Client};
 use log::{debug, info};
 use rand::Rng;
-use serde_json::{from_str, to_string, Value};
+use serde_json::Value;
 
 pub type Result<T, E = ConductorError> = std::result::Result<T, E>;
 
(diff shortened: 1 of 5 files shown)


conductor/src/lib.rs
```diff
@@ -290,56 +290,102 @@ pub async fn delete_namespace(client: Client, name: &str) -> Result<(), Conducto
     Ok(())
 }
 
-async fn get_secret_for_db(client: Client, name: &str) -> Result<Secret, ConductorError> {
+async fn get_secret_for_db(client: Client, name: &str) -> Result<(Secret, Secret), ConductorError> {
```
sjmiller609, line 293: Does this work instead of doing Box for AwsError?
```
Result<(Secret, Secret), Box<ConductorError>>
```

conductor/src/types.rs
```diff
@@ -40,7 +40,7 @@ pub struct StateToControlPlane {
     pub event_type: Event,     // pass through from event that triggered a data plane action
     pub spec: Option<CoreDBSpec>,
     pub status: Option<CoreDBStatus>,
-    pub connection: Option<types::ConnectionInfo>,
+    pub connection: Option<(types::ConnectionInfo, types::ConnectionInfo)>,
```
sjmiller609, line 43: In order to keep things backwards compatible so that these new messages can be received by the control plane properly, do something like this instead:

```
pub struct ConnectionInfo {
    pub host: String,
    pub port: u16,
    pub user: String,
    pub password: String,
    pub app_user: String,
    pub app_password: String,
}
```

This can serialize and include the new information, and it still will deserialize the same, with the app user information ignored until handled on control plane size

conductor/src/lib.rs
sjmiller609: Example https://github.com/tembo-io/tembo-stacks/blob/d4e7288f45aff7923d35f53bba06caac37f915b1/conductor/src/lib.rs#L53

sjmiller609: No actually I think you did it the correct way, and I did the not recommended approach https://github.com/rust-lang/rust-clippy/pull/9373 so nevermind

sjmiller609 (changes requested):

conductor/src/lib.rs
nhudson: Yeah I spent some time fighting the compiler on this one.  It wasn't something I had seen before.

conductor/src/types.rs
nhudson: This should be resolved now.

sjmiller609 (approved):

PR #223, opened 2023-08-23
Ensure extensions are installed in all psql pods

nhudson:
With the ability to enable HA/replicas, we need to make sure when we trunk install extensions that they are installed to all replicas and not just to the primary.

fixes: [TEM-1620](https://linear.app/tembo/issue/TEM-1620/handle-extension-installation-into-replicas)


sjmiller609 (changes requested):
On this section of reconcile_installs
```
    // Get extensions in spec.trunk_install that are not in status.trunk_install
    let trunk_installs = cdb
        .spec
        .trunk_installs
        .iter()
        .filter(|&ext| {
            !cdb.status
                .clone()
                .unwrap_or_default()
                .trunk_installs
                .unwrap_or_default()
                .iter()
                .any(|ext_status| ext.name == ext_status.name)
        })
        .collect::<Vec<_>>();
    install_extensions(cdb, trunk_installs, ctx.clone()).await

```
    
It checks if something has already been trunk installed, only performing the install where it hasn't been performed yet. I think could have a problem when switching number of replicas:

- Install pg_jsonschema
- Increase replicas to 2
- Now trunk install is already in status.trunk_installs, but not on the second replica - so it never gets installed


Here is my thought on how it could be accomplished:

In TrunkInstallStatus, add a new field:

```
struct TrunkInstallStatus {
    ...
    installed_to_instances: Option<Vec<String>>
}
```

- In install_extensions, now require a pod name argument, and only do it for one pod 
- In reconcile_installs, use the new field to determine the a list of (trunk install, pod name) instead of a list of trunk installs.
- Pass the pod name to install_extensions from reconcile_installs
- move add_trunk_install_to_status out from install_extension, into reconcile_installs so that it can handle setting the installed_to_instances field

tembo-operator/src/extensions/types.rs
```diff
@@ -14,6 +14,7 @@ pub struct TrunkInstallStatus {
     pub version: Option<String>,
     pub error: bool,
     pub error_message: Option<String>,
+    pub installed_to_instances: Vec<String>,
```
sjmiller609, line 17: I'm not sure maybe we need `Option<Vec<String>>` to handle existing instances when the CRD updates

tembo-operator/src/extensions/install.rs
```diff
@@ -8,14 +9,15 @@ use crate::{
...
-
-use crate::apis::coredb_types::CoreDBStatus;
+use tracing::{debug, error, info, instrument};
 
 pub async fn reconcile_trunk_installs(
     cdb: &CoreDB,
     ctx: Arc<Context>,
 ) -> Result<Vec<TrunkInstallStatus>, Action> {
+    // Fetch all pods
+    let all_pods = cdb.pods_by_cluster(ctx.client.clone()).await?;
```
sjmiller609, line 19: I think this should be called right before it's used, so it's least likely that something changed since calling

tembo-operator/src/extensions/install.rs
```diff
@@ -76,14 +78,99 @@ pub async fn reconcile_trunk_installs(
                 .any(|ext_status| ext.name == ext_status.name)
         })
         .collect::<Vec<_>>();
-    install_extensions(cdb, trunk_installs, ctx.clone()).await
+
```
sjmiller609, line 81: I think the `let trunk_installs =...` line above needs adjustment to generate the trunk installs list as (this extension, this pod name) instead of (this extension)

tembo-operator/src/extensions/install.rs
```diff
@@ -76,14 +78,99 @@ pub async fn reconcile_trunk_installs(
...
+            error_message: Some(output),
+            installed_to_instances: vec![pod_name.to_string()],
+        };
+        add_trunk_install_to_status(coredb_api, coredb_name, &trunk_install_status).await?;
+        Ok(trunk_install_status)
+    }
 }
 
 /// handles installing extensions
+#[instrument(skip(ctx, cdb))]
```
sjmiller609, line 168: nice

tembo-operator/src/extensions/types.rs
nhudson: Yeah I noticed that too.  I will get that added before I wrap this up

PR #224, opened 2023-08-24
Use ExtensionStatus type when determining available extensions

sjmiller609:
This is a small change and just here to help with a following change that handles the case of loading libraries for extensions. Since this part can go independently, I have in separate PR.


tembo-operator/src/extensions/database_queries.rs
```diff
@@ -152,21 +152,23 @@ pub fn parse_databases(psql_str: &str) -> Vec<String> {
 }
 
 /// list databases then get all extensions from each database
-pub async fn get_all_extensions(cdb: &CoreDB, ctx: Arc<Context>) -> Result<Vec<Extension>, Action> {
+pub async fn get_all_extensions(cdb: &CoreDB, ctx: Arc<Context>) -> Result<Vec<ExtensionStatus>, Action> {
```
sjmiller609, line 155: When getting all extensions from the system, we can use the ExtensionStatus type. That will allow us to add details about the extension category, like requires create extension true / false and requires load true / false

nhudson (approved):

PR #227, opened 2023-08-28
Enabling extensions that require libraries to be loaded ahead of time

sjmiller609:


tembo-operator/src/extensions/database_queries.rs
```diff
@@ -215,12 +374,24 @@ fn generate_extension_enable_cmd(
...
+        None => {
+            error!("There should always be an extension status before attempting to toggle an extension");
+            return Err("Extension is not installed".to_string());
+        }
+        Some(status) => status
+    };
+    if !current_status.create_extension {
+        // If the extension does not require CREATE EXTENSION, then we do not need to do anything in this function.
+        return Ok(());
+    }
```
sjmiller609, line 393: instead of changing this function, make a new function for setting shared_preload_libraries for anything we will enable before going into here

tembo-operator/src/extensions/toggle.rs
```diff
@@ -40,37 +41,37 @@ async fn toggle_extensions(
     let mut ext_status_updates = ext_status_updates.clone();
     for extension_to_toggle in toggle_these_extensions {
         for location_to_toggle in extension_to_toggle.locations {
-            match database_queries::toggle_extension(
+            match database_queries::load_extension_if_required(
```
sjmiller609, line 44: move this to a separate function called before toggle extension to ensure we set all libraries at once then wait for restart to complete before attempting any toggle extensions

tembo-operator/src/apis/coredb_types.rs
```diff
@@ -99,66 +103,95 @@ pub struct CoreDBSpec {
...
-        // e.g. stack defines shared_preload_libraries = pg_cron, then operator installs pg_stat_statements at runtime
-        // we need to merge the two configs into one,  shared_preload_libraries = pg_cron, pg_stat_statements
-        let mut merged_multival_configs: Vec<PgConfig> = Vec::new();
-        for cfg_name in MULTI_VAL_CONFIGS {
-            let merged_config = merge_pg_configs(&stack_configs, &runtime_configs, cfg_name)?;
-            if let Some(merged_config) = merged_config {
-                merged_multival_configs.push(merged_config);
+// extracts all postgres configurations
+// configs can be defined in several different places (from a stack, user override, from an extension installation, user overrides, etc)
+pub fn get_pg_configs(cdb: &CoreDB) -> Result<Option<Vec<PgConfig>>, MergeError> {
```
sjmiller609, line 108: instead of being impl on CoreDBSpec, it now accepts a whole cdb so it can also look at status. Since the status is needed to know if we can enable shared_preload_libraries

tembo-operator/src/extensions/database_queries.rs
```diff
@@ -1,24 +1,54 @@
...
 use regex::Regex;
-use std::{collections::HashMap, sync::Arc};
+use std::{collections::HashMap, sync::Arc, time::Duration};
 use tracing::{debug, error, info, warn};
 
 lazy_static! {
     static ref VALID_INPUT: Regex = Regex::new(r"^[a-zA-Z]([a-zA-Z0-9]*[-_]?)*[a-zA-Z0-9]+$").unwrap();
 }
 
+// TODO: Get this list from trunk instead of coding it here
```
sjmiller609, line 18: https://github.com/tembo-io/trunk/pull/407

sjmiller609: I want to check this PR is working OK then do another PR for the fetch part. I want to have some way to do it but tolerant to external query failing, like fetch and write to configmap, then only update sometimes. but trying to keep the total code change in this pr smaller

sjmiller609:
operator tests flaked. checking on that

tembo-operator/src/extensions/database_queries.rs
sjmiller609: starting this in new branch now

PR #228, opened 2023-08-30
Fix issue with listing current shared_preload_libraries

sjmiller609:


tembo-operator/src/extensions/toggle.rs
```diff
@@ -17,10 +17,7 @@ use crate::{
     },
 };
 use std::{sync::Arc, time::Duration};
-use tracing::{
-    error,
-    log::{debug, info},
-};
```
sjmiller609, line 23: logs imported from tracing::log::* do not show up, we should use tracing::*

nhudson (approved):

tembo-operator/src/extensions/toggle.rs
nhudson: yes you should be using the `tracing` crate.  It should be `use tracing::{error, debug, info};`

PR #238, opened 2023-09-05
Check for library to be ready before enabling an extension

sjmiller609:


tembo-operator/src/exec.rs
```diff
@@ -99,10 +99,11 @@ impl ExecCommand {
                     result_stdout.clone(),
                     result_stderr.clone()
                 );
-                error!(
-                    "Error executing command: {:?} on pod: {:?}. response: {:?}",
-                    command, self.pod_name, output
+                warn!(
+                    "Error executing command on pod: {:?}. response: {:?}",
+                    self.pod_name, output
```
sjmiller609, line 104: it's not always an error when the command returns an error - for example if the DB is shutting down while we try to run a psql.

tembo-operator/src/extensions/toggle.rs
```diff
@@ -11,11 +11,40 @@ use kube::runtime::controller::Action;
...
         kubernetes_queries::merge_location_status_into_extension_status_list, types::get_location_status,
     },
 };
-use std::sync::Arc;
-use tracing::error;
+use std::{sync::Arc, time::Duration};
+use tracing::{error, warn};
+
+
+// TODO: @sjmiller609 fetch from trunk
```
sjmiller609, line 22: I have this in progress in a different PR, but trying to minimize the code volume in any given pr

tembo-operator/src/extensions/database_queries.rs
```diff
@@ -132,11 +165,11 @@ pub async fn list_databases(cdb: &CoreDB, ctx: Arc<Context>) -> Result<Vec<Strin
         .psql(LIST_DATABASES_QUERY.to_owned(), "postgres".to_owned(), ctx)
         .await?;
     let result_string = psql_out.stdout.unwrap();
-    Ok(parse_databases(&result_string))
+    Ok(parse_sql_output(&result_string))
 }
 
-pub fn parse_databases(psql_str: &str) -> Vec<String> {
-    let mut databases = vec![];
+pub fn parse_sql_output(psql_str: &str) -> Vec<String> {
```
sjmiller609, line 171: this function can just parse any single-column return value

ChuckHend (approved):

PR #239, opened 2023-09-06
Auto-enable libraries for extensions that require load

sjmiller609:
This PR covers extensions that require load and require create extension, but not extensions that require load and not create extension.

Allows enabling these extensions
```
citus
pg_cron
pg_later
pglogical
pglogical_ticker
pg_net
pg_tle
plrust
timescaledb
```

### DIFF
--- a/tembo-operator/src/apis/coredb_types.rs
+++ b/tembo-operator/src/apis/coredb_types.rs
@@ -16,7 +16,9 @@ use kube::CustomResource;
 use crate::extensions::types::{Extension, TrunkInstall, TrunkInstallStatus};
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, BTreeSet};
+use crate::apis::postgres_parameters::ConfigValue;
+use crate::extensions::toggle::REQUIRES_LOAD;
 
 #[derive(Deserialize, Serialize, Clone, Debug, JsonSchema, Default)]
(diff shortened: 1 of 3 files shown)


tembo-operator/tests/integration_tests.rs
```diff
@@ -1735,4 +1735,137 @@ mod test {
         // Delete namespace
         let _ = delete_namespace(client.clone(), &namespace).await;
     }
+
+    #[tokio::test]
+    #[ignore]
+    async fn functional_test_shared_preload_libraries() {
```
sjmiller609, line 1741: this test checks we can turn on pg_cron and citus

tembo-operator/src/apis/coredb_types.rs
```diff
@@ -108,12 +114,52 @@ impl CoreDBSpec {
...
             .and_then(|s| s.postgres_config.clone())
             .unwrap_or_default();
-        let runtime_configs = self.runtime_config.clone().unwrap_or_default();
+        let mut runtime_configs = self.runtime_config.clone().unwrap_or_default();
         // TODO: configs that come with extension installation
         // e.g. let extension_configs = ...
         // these extensions could be set by the operator, or trunk + operator
         // trunk install pg_partman could come with something like `pg_partman_bgw.dbname = xxx`
 
+        // Get list of extension names that require load
```
sjmiller609, line 123: the following code includes to shared_preload_libraries any extension name that requires load and is enabled in spec

ianstanton (approved):

PR #240, opened 2023-09-06
Update metrics to allow 7 day queries

sjmiller609:


sjmiller609:
front end does not need this yet

PR #245, opened 2023-09-12
Fetch metadata from trunk

sjmiller609:
use this endpoint to determine the list of extensions that require load https://registry.pgtrunk.io/extensions/libraries


sjmiller609:
just did merge from main, need to clean up some differences

ChuckHend (approved):

PR #247, opened 2023-09-15
Add Stack types and config engines

ChuckHend:
Adds Stack and their config engines.

Operator does not consume them yet, so should be no change to any operator behaviors. Purposely leaving the operator on its own "Stack" struct here: https://github.com/tembo-io/tembo-stacks/blob/81e00154c45b4608358f8d7119449331f78c954c/tembo-operator/src/apis/coredb_types.rs#L238-L242, separate from what we're moving over from our control-plane


tembo-operator/Cargo.toml
```diff
@@ -1,10 +1,10 @@
 [package]
 name = "controller"
-version = "0.0.1"
+version = "0.0.2-alpha"
 edition = "2021"
 default-run = "controller"
 license = "Apache-2.0"
-publish = false
+publish = true
```
ChuckHend, line 7: start publishing this crate

tembo-operator/src/apis/postgres_parameters.rs
```diff
@@ -204,6 +205,18 @@ impl From<WrapValue> for Result<ConfigValue, JsonParsingError> {
     }
 }
 
+// TODO: NOT SURE IF THIS WORKS
+impl From<&str> for ConfigValue {
+    fn from(item: &str) -> Self {
+        let values: Vec<String> = item.split(',').map(|s| s.trim().to_string()).collect();
+        if values.len() > 1 {
+            ConfigValue::Multiple(values.into_iter().collect())
```
ChuckHend, line 213: this part might be wrong, still need to test it

.github/workflows/cargo-test.yaml
```diff
@@ -74,7 +74,7 @@ jobs:
     runs-on:
       - self-hosted
       - dind
-      - small
+      - large-8x8
```
ChuckHend, line 77: was running out of memory in CI

tembo-operator/src/extensions/types.rs
```diff
@@ -38,13 +38,13 @@ impl Default for Extension {
...
 
-#[derive(Clone, Debug, Deserialize, Eq, Hash, JsonSchema, Serialize, PartialEq)]
+#[derive(Clone, Debug, Deserialize, Eq, Hash, JsonSchema, Serialize, PartialEq, ToSchema)]
 pub struct ExtensionInstallLocation {
     pub enabled: bool,
     #[serde(default = "defaults::default_database")]
     pub database: String,
     pub version: Option<String>,
-    schema: Option<String>,
+    pub schema: Option<String>,
```
ChuckHend, line 47: @sjmiller609 - do you remember why schema was made private, while rest of the attributes are public?

sjmiller609: Yes. I wanted to make sure it was not accessible outside the module to other parts while removing the incorrect schema logic, but that's ok to turn it back on now that it's working fine

sjmiller609 (approved):

PR #249, opened 2023-09-15
add action and workflow to publish controller

ChuckHend:
The workflow should identify any crate in the repo, so setting every Cargo.toml to `publish = false` except for the tembo-operator.

### DIFF
--- a/tembo-operator/Cargo.toml
+++ b/tembo-operator/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "controller"
 description = "Tembo Operator for Postgres"
-version = "0.1.0"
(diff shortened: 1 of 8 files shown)


tembo-operator/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "controller"
 description = "Tembo Operator for Postgres"
-version = "0.1.0"
+version = "0.1.1"
```
ChuckHend, line 4: bumping minor here so that the workflow is triggered when this merges

ianstanton (approved):

PR #251, opened 2023-09-15
Don't CASCADE if toggling off extension

sjmiller609:
Instead, this will trigger error message if there are objects to be dropped


sjmiller609:
we decided leave it how it is.

PR #252, opened 2023-09-15
stack ver optional

ChuckHend:


tembo-operator/src/stacks/types.rs
```diff
@@ -88,6 +88,10 @@ impl Stack {
     }
 }
 
+fn default_stack_image() -> Option<String> {
+    Some(default_image())
```
ChuckHend, line 92: wrapper needed because operator default_image does returns a string

sjmiller609 (approved):

PR #254, opened 2023-09-18
replace coredb Stack w/ Stacks def

ChuckHend:
Use the Stacks struct from the `Stacks` API instead of coredb types, which requires update to the crd. It is a non-breaking change because all the new attributes on `Stack` are optional.

### DIFF
--- a/tembo-operator/src/stacks/types.rs
+++ b/tembo-operator/src/stacks/types.rs
@@ -1,6 +1,6 @@
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
-use std::{collections::BTreeMap, default};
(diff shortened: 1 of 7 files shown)


tembo-operator/src/stacks/types.rs
```diff
@@ -11,12 +11,6 @@ use crate::{
     stacks::config_engines::{olap_config_engine, standard_config_engine, ConfigEngine},
 };
 
-#[derive(Clone, Debug, Serialize, Deserialize, JsonSchema)]
```
ChuckHend, line 14: this was a placeholder

nhudson (approved):

PR #257, opened 2023-09-18
Add Connection Pooling

ianstanton:
Allow for enable / disable and basic configuration of connection pooler (PgBouncer). There will be follow-up PRs related to this feature.


charts/tembo-operator/templates/crd.yaml
```diff
@@ -49,6 +49,3935 @@ spec:
                     nullable: true
                     type: string
                 type: object
+              connPooler:
```
ianstanton, line 52: We want to see if we can avoid adding all of this. Most lines are likely coming from `PodTemplateSpec`

ianstanton:
Seeing test failures related to https://github.com/tembo-io/tembo-stacks/commit/9b41c9dbf5d25a472325adef540c9e2f292abb39 in main

tembo-operator/src/cloudnativepg/cnpg.rs
```diff
@@ -916,6 +917,72 @@ pub async fn reconcile_cnpg(cdb: &CoreDB, ctx: Arc<Context>) -> Result<(), Actio
...
+    let client = ctx.client.clone();
+    let name = pooler
+        .metadata
+        .name
+        .clone()
+        .expect("Pooler should always have a name");
+    let namespace = pooler
+        .metadata
+        .namespace
+        .clone()
```
vrmiguel, line 972:
```suggestion
        .as_ref()
```

:rocket:

tembo-operator/src/defaults.rs
```diff
@@ -121,6 +122,26 @@ pub fn default_backup_schedule() -> Option<String> {
...
+pub fn default_conn_pooler_enabled() -> bool {
+    false
+}
+
+pub fn default_pgbouncer() -> PgBouncer {
+    PgBouncer {
+        poolMode: PoolerPgbouncerPoolMode::Transaction,
+        parameters: None,
+        // TODO(ianstanton): Add some sensible defaults for pgbouncer resources
+        resources: None,
```
ianstanton, line 141: Resources aren't passed through yet. That will be handled in a follow-up.

tembo-operator/yaml/sample-with-pooler.yaml
```diff
@@ -0,0 +1,33 @@
...
+      value: top
+  connPooler:
+    enabled: true
+    pooler:
+      poolMode: transaction
+      # Valid parameter values can be found at https://www.pgbouncer.org/config.html
+      parameters:
+        default_pool_size: "50"
+        max_client_conn: "5000"
+      resources: ~
```
sjmiller609, line 33: I think the spec looks nice with good balance of config and defaults

Two nits:
- "connPooler" -> instead "connectionPooler" or "pooler"
- why connPooler + pooler instead of flatter like this:

```
    pooler:
      enabled: true
      poolMode: transaction
      # Valid parameter values can be found at https://www.pgbouncer.org/config.html
      parameters:
        default_pool_size: "50"
        max_client_conn: "5000"
      resources: ~
```

sjmiller609 (approved):

vrmiguel (approved):

tembo-operator/yaml/sample-with-pooler.yaml
ianstanton: I agree, will change `connPooler` to `connectionPooler`. I was thinking the additional layer would make it easier for us to add support for other poolers in the future. Maybe we change `pooler` to `pgbouncer`? If we add support for something like `pgcat` in the future, we could just add a new struct for it. I think we'd need to move `enabled` into each layer for this. Something like:
```
connectionPooler:
  pgbouncer:
    enabled: false
    poolMode: transaction
    parameters:
      default_pool_size: "50"
      max_client_conn: "5000"
  pgcat:
    enabled: true
    someConfig: value
    anotherConfig: value
```
@sjmiller609 WDYT?

tembo-operator/src/apis/coredb_types.rs
```diff
@@ -102,6 +103,24 @@ pub struct Restore {
...
+    pub pooler: PgBouncer,
+}
+
+#[derive(Deserialize, Serialize, Clone, Debug, JsonSchema, Default)]
+#[allow(non_snake_case)]
+pub struct PgBouncer {
+    pub poolMode: PoolerPgbouncerPoolMode,
+    // Valid parameter values can be found at https://www.pgbouncer.org/config.html
+    pub parameters: Option<BTreeMap<String, String>>,
+    pub resources: Option<ResourceRequirements>,
```
ChuckHend, line 121: when we support more than just PgBouncer, will the other pools also have `parameters` and `resources`? If so, I think these should live under `ConnPooler`, but if they are unique to PgBouncer,then I think it makes sense to live here. 

#[derive(Deserialize, Serialize, Clone, Debug, JsonSchema, Default)]
#[allow(non_snake_case)]
pub struct ConnPooler {
    #[serde(default = "defaults::default_conn_pooler_enabled")]
    pub enabled: bool,
    #[serde(default = "defaults::default_pgbouncer")]
    pub pooler: PgBouncer,
    pub parameters: Option<BTreeMap<String, String>>,
    pub resources: Option<ResourceRequirements>,
}

ChuckHend (approved):

tembo-operator/src/apis/coredb_types.rs
```diff
@@ -102,6 +103,24 @@ pub struct Restore {
     pub s3_credentials: Option<S3Credentials>,
 }
 
+#[derive(Deserialize, Serialize, Clone, Debug, JsonSchema, Default)]
+#[allow(non_snake_case)]
+pub struct ConnPooler {
+    #[serde(default = "defaults::default_conn_pooler_enabled")]
+    pub enabled: bool,
+    #[serde(default = "defaults::default_pgbouncer")]
+    pub pooler: PgBouncer,
```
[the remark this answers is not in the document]
ianstanton: I ran into some issues with defaults, so will follow up with this

PR #260, opened 2023-09-19
update OLAP configurations

ChuckHend:


ianstanton (approved):

tembo-operator/src/stacks/templates/olap.yaml
```diff
@@ -15,16 +15,30 @@ compute_templates:
...
     value: 30min
-  - name: pg_stat_statements.track
-    value: all
-  - name: track_io_timing
-    value: 'on'
   - name: cron.host
     value: /controller/run
+  - name: log_min_duration_statement
+    value: 10000
+  - name: pg_stat_statements.track
```
samay-sharma, line 30: Make this default i.e. top

tembo-operator/src/stacks/config_engines.rs
```diff
@@ -64,32 +66,80 @@ pub fn standard_config_engine(stack: &Stack) -> Vec<PgConfig> {
...
+    i32::max((cpu as f64 * 0.5).floor() as i32, 2)
+}
+
+fn olap_max_parallel_workers(cpu: i32) -> i32 {
+    // higher of the default (8) or 2 * cpu
+    i32::max(2 * cpu, 8)
+}
+
+// returns work_mem value in MB
+fn olap_work_mem(sys_mem_mb: i32, shared_buffers_mb: i32, max_connections: i32) -> i32 {
```
samay-sharma, line 129: Formula for work_mem should be same, right? By changing max_connections, shouldn't that adjust work_mem correctly.

This is just asking, not yet a suggestion.

tembo-operator/src/stacks/templates/olap.yaml
```diff
@@ -15,16 +15,30 @@ compute_templates:
...
+    value: all
+  - name: random_page_cost
+    value: 1.5
+  - name: seq_page_cost
```
samay-sharma, line 34: This is default, why do we need to call it out?

tembo-operator/src/stacks/templates/olap.yaml
```diff
@@ -15,16 +15,30 @@ compute_templates:
...
-    value: all
-  - name: track_io_timing
-    value: 'on'
   - name: cron.host
     value: /controller/run
+  - name: log_min_duration_statement
+    value: 10000
+  - name: pg_stat_statements.track
+    value: all
+  - name: random_page_cost
```
samay-sharma, line 32: I'm wondering if this should be a higher value in this case, like maybe 2 (just because olap won't be using as many indexes due to large scans). What's our default for OLTP?

tembo-operator/src/stacks/config_engines.rs
ChuckHend: Hmm, yes I think you're right.

tembo-operator/src/stacks/templates/olap.yaml
ChuckHend: We use the default in OLTP. MQ stack uses 1.1 though.

tembo-operator/src/stacks/config_engines.rs
```diff
@@ -64,32 +66,80 @@ pub fn standard_config_engine(stack: &Stack) -> Vec<PgConfig> {
...
+        },
+        PgConfig {
+            name: "work_mem".to_owned(),
+            value: ConfigValue::Single(format!("{work_mem}MB")),
         },
     ]
 }
 
+// olap formula for max_parallel_workers_per_gather
+fn olap_max_parallel_workers_per_gather(cpu: i32) -> i32 {
```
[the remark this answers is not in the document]
ChuckHend: changed to min of 2, or 8. scaled to 0.5* cpu in between those two values

tembo-operator/src/stacks/config_engines.rs
```diff
@@ -64,32 +66,80 @@ pub fn standard_config_engine(stack: &Stack) -> Vec<PgConfig> {
...
+    // higher of default (2) or 0.5 * cpu
+    i32::max((cpu as f64 * 0.5).floor() as i32, 2)
+}
+
+fn olap_max_parallel_workers(cpu: i32) -> i32 {
```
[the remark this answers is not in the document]
ChuckHend: updated both

PR #262, opened 2023-09-20
slim stack

ChuckHend:


tembo-operator/src/apis/coredb_types.rs
```diff
@@ -1,4 +1,4 @@
-use crate::{extensions::types::ExtensionStatus, stacks::types::Stack};
```
ChuckHend, line 1: use slim Stack struct as a placeholder until we do a larger refactor

tembo-operator/src/apis/coredb_types.rs
```diff
@@ -21,6 +21,13 @@ use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 use std::collections::{BTreeMap, BTreeSet};
 
+#[derive(Clone, Default, Debug, Serialize, Deserialize, JsonSchema)]
+pub struct Stack {
```
ChuckHend, line 25: ^ slim Stack

PR #264, opened 2023-09-20
Add secrets service

sjmiller609:
Our ingress does authentication based on the namespace path, and also there is a network policy to only allow network access from our ingress namespace into this service.


dataplane-webserver/src/main.rs
```diff
@@ -70,6 +71,11 @@ async fn main() -> std::io::Result<()> {
             .wrap(middleware::Logger::default())
             .service(web::scope("/").service(root::ok))
             .service(web::scope("/{namespace}/metrics").service(metrics::query_range))
+            .service(
+                web::scope("/{namespace}/secrets")
+                    .service(secrets::get_secret)
+                    .service(secrets::get_secret_names),
+            )
```
sjmiller609, line 78: needs open api doc configuration

nhudson (approved):

ianstanton (approved):

PR #267, opened 2023-09-21
Create, Update, Delete App Deployments

ChuckHend:
creates a Kubernetes Deployment for every AppService in `spec.app_services`


tembo-operator/src/app_service/manager.rs
```diff
@@ -0,0 +1,239 @@
...
+                    port: IntOrString::String(probes.liveness.port),
+                    ..HTTPGetAction::default()
+                }),
+                initial_delay_seconds: Some(probes.liveness.initial_delay_seconds as i32),
+                ..Probe::default()
+            };
+            (Some(readiness_probe), Some(liveness_probe))
+        }
+        None => {
+            // are there default probes we could configure when none are provided?
```
ianstanton, line 69: We could use a TCP `livenessProbe` by default https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-tcp-liveness-probe. I'm not aware of a default `readinessProbe`

tembo-operator/src/app_service/manager.rs
```diff
@@ -0,0 +1,239 @@
...
+        env.into_iter()
+            .map(|(k, v)| EnvVar {
+                name: k,
+                value: Some(v),
+                ..EnvVar::default()
+            })
+            .collect()
+    });
+    // TODO: Container VolumeMounts, currently not in scope
+    // TODO: PodSpec volumes, currently not in scope
```
ianstanton, line 108: Are we planning to add a `Service` for each `Deployment` as a follow-up? If not, I think we should.

tembo-operator/tests/integration_tests.rs
```diff
@@ -2579,6 +2579,195 @@ mod test {
...
+            "limits": {
+                "cpu": "50m",
+                "memory": "128Mi"
+            }
+        }))
+        .unwrap();
+        let app_1_resources = app_1_container.resources.unwrap();
+        assert_eq!(app_1_resources, expected);
+
+
```
ianstanton, line 2770: Let's drop one of the appServices and assert only one expected Deployment is present

tembo-operator/src/app_service/manager.rs
```diff
@@ -0,0 +1,239 @@
...
+        match deployment_api
+            .patch(&ad.name, &ps, &Patch::Apply(&ad.deployment))
+            .await
+            .map_err(Error::KubeError)
+        {
+            Ok(_) => {
+                debug!("Successfully reconciled AppService: {}", ad.name);
+            }
+            Err(e) => {
+                // is here a better way to surface failures without completely stopping all reconciliation of AppService?
```
ianstanton, line 232: I think we can take the approach we use in other reconcile functions and requeue:
https://github.com/tembo-io/tembo-stacks/blob/c7e146d56c89c87eb02030719bf59a9bd30fd167/tembo-operator/src/cloudnativepg/cnpg.rs#L636-L639

tembo-operator/yaml/sample-coredb.yaml
```diff
@@ -21,4 +21,4 @@ spec:
     - name: shared_preload_libraries
       value: 'pg_cron,pg_stat_statements'
     - name: pg_stat_statements.track
-      value: top
\ No newline at end of file
+      value: top
```
ianstanton, line 24: Eventually, it'd be nice to have a sample with `appServices` either in this file or a new one

ianstanton (approved):
Left some comments, but it's looking like a good v1!

tembo-operator/src/apis/coredb_types.rs
```diff
@@ -109,6 +110,8 @@ pub struct CoreDBSpec {
     pub runtime_config: Option<Vec<PgConfig>>,
     // configuration overrides, typically defined by the user
     pub override_configs: Option<Vec<PgConfig>>,
+
+    pub app_services: Option<Vec<AppService>>,
```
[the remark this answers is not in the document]
ChuckHend: Rust convention is snake case. 

We can add the macro to this that will cast it to camel case externally though, then both are compliant.

tembo-operator/src/app_service/types.rs
```diff
@@ -0,0 +1,107 @@
...
+    pub args: Option<Vec<String>>,
+    pub command: Option<Vec<String>>,
+    pub env: Option<BTreeMap<String, String>>,
+    // PortMapping is in format of String "host:container"
+    pub ports: Option<Vec<PortMapping>>,
+    pub resources: Option<ResourceRequirements>,
+    pub probes: Option<Probes>,
+    pub metrics: Option<Metrics>,
+    pub ingress: Option<IngressRouteTCPSpec>,
+}
```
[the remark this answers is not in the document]
ChuckHend: Ok

tembo-operator/src/app_service/manager.rs
ChuckHend: Cool I will set that up.

tembo-operator/yaml/sample-coredb.yaml
ChuckHend: Will definitely do that when the feature is ready to use.

tembo-operator/src/app_service/manager.rs
ChuckHend: Is that how you want to handle this? If we fail to reconcile a single AppService, that we stop trying on the rest of the AppService?

Would that mean that if we do that here, then we would skip reconciling all the other resources below this line? https://github.com/tembo-io/tembo-stacks/blob/c05689c41b8caa8c6ddf9d198653d72f3b52e589/tembo-operator/src/controller.rs#L170

tembo-operator/src/apis/coredb_types.rs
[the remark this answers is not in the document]
ianstanton: That would be perfect, hadn't thought of that

tembo-operator/src/app_service/manager.rs
```diff
@@ -0,0 +1,239 @@
...
+    }
+    if to_delete.is_empty() {
+        None
+    } else {
+        Some(to_delete)
+    }
+}
+
+
+pub async fn reconcile_app_services(cdb: &CoreDB, ctx: Arc<Context>) -> Result<(), Error> {
```
sjmiller609, line 178: Here we can return Result<(), Action>, then return requeues that are long (300-600s) for unexpected issues and have error log. For expected errors, like if there is a wait condition, then requeue short at a minimum of 5s

tembo-operator/src/app_service/manager.rs
ChuckHend: Actually - @ianstanton do you mind if we handle that when we implement Ingress? Thats next.

tembo-operator/src/app_service/manager.rs
```diff
@@ -0,0 +1,239 @@
...
+        Some(appsvcs) => appsvcs,
+        None => {
+            debug!("No AppServices found in Instance: {}", ns);
+            return Ok(());
+        }
+    };
+
+    let deployments: Vec<AppDeployment> = appsvcs
+        .iter()
+        .map(|appsvc| generate_deployment(appsvc, &ns, oref.clone()))
```
sjmiller609, line 217: A couple of things to watch out for:
- multiple CoreDB in same namespace with same deployment names should work, so might need cdb name in deployment name (maybe I missed where it comes from)
- too long resource names

tembo-operator/src/controller.rs
```diff
@@ -166,6 +167,11 @@ impl CoreDB {
             }
         };
 
+        reconcile_app_services(self, ctx.clone()).await.map_err(|e| {
+            error!("Error reconciling AppService deployment: {:?}", e);
```
sjmiller609, line 171: I like to be able to grep all logs by CoreDB name, where possible

tembo-operator/src/controller.rs
```diff
@@ -166,6 +167,11 @@ impl CoreDB {
...
+            Action::requeue(Duration::from_secs(5))
```
sjmiller609, line 172: Mentioned in other comment, I think there are some unexpected errors that we should requeue for longer

sjmiller609 (approved):
A few feedbacks, looks good

tembo-operator/src/apis/coredb_types.rs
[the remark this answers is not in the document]
ChuckHend: @ianstanton - I pushed that change. Going to resolve this conversation.

tembo-operator/tests/integration_tests.rs
ChuckHend: Added delete tests; delete one then delete all AppService

tembo-operator/src/app_service/manager.rs
ianstanton: Yep just wanted to make sure it was on the list of things to follow up with

tembo-operator/src/controller.rs
ChuckHend: I think we'll run into name length constraints pretty quick, given these pods have the hash at the end?

I was running into issues with name length of the coredbs we have in the sample yaml.

ChuckHend: Oh I thought our tracing system was handling that now? If not I'll update this log.

tembo-operator/src/app_service/manager.rs
```diff
@@ -0,0 +1,262 @@
...
+};
+use std::{collections::BTreeMap, sync::Arc, time::Duration};
+
+use super::types::AppService;
+use tracing::*;
+
+// private wrapper to hold the AppService and the name of the Deployment
+struct AppDeployment {
+    deployment: Deployment,
+    name: String,
```
sjmiller609, line 27: To clarify issue with name:

For example, this config of two coredb in same namespace:

```
---
apiVersion: coredb.io/v1alpha1
kind: CoreDB
metadata:
  name: test-coredb
  namespace: default
spec:
  app_services:
  - name: postgrest
     .... etc
---
apiVersion: coredb.io/v1alpha1
kind: CoreDB
metadata:
  name: test-coredb-2
  namespace: default
spec:
  app_services:
  - name: postgrest
     .... etc
```

But the kubernetes convention is that resources never collide when they have different names, and they should be deployable in same namespace.

tembo-operator/src/app_service/types.rs
```diff
@@ -0,0 +1,107 @@
...
+    pub image: String,
+    pub args: Option<Vec<String>>,
+    pub command: Option<Vec<String>>,
+    pub env: Option<BTreeMap<String, String>>,
+    // PortMapping is in format of String "host:container"
+    pub ports: Option<Vec<PortMapping>>,
+    pub resources: Option<ResourceRequirements>,
+    pub probes: Option<Probes>,
+    pub metrics: Option<Metrics>,
+    pub ingress: Option<IngressRouteTCPSpec>,
```
sjmiller609, line 20: I think here we should define our own struct not use IngressRouteTCPSpec

sjmiller609 (changes requested):

charts/tembo-operator/templates/crd.yaml
```diff
@@ -25,6 +25,227 @@ spec:
...
+                      nullable: true
+                      type: array
+                    env:
+                      additionalProperties:
+                        type: string
+                      nullable: true
+                      type: object
+                    image:
+                      type: string
+                    ingress:
```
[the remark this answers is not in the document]
sjmiller609: redundant comment with below review comment

sjmiller609 (approved):

PR #268, opened 2023-09-21
Expose first recovery point time to CoreDB status

nhudson:
First change needed to add PITR 

This change adds `first_recoverability_time` field to the `CoreDBStatus`.  This exposes the earliest full backup we could potentially restore to and will be needed to calculate the dates and times in the UI to do a PITR.

This also removes unneeded telemetry spans that are already in place with the `#[instrument]` macros.

fixes: [TEM-1809](https://linear.app/tembo/issue/TEM-1809/add-first-recoverability-point-to-status)

### DIFF
--- a/tembo-operator/justfile
+++ b/tembo-operator/justfile
@@ -75,6 +75,7 @@ start-kind:
 	just update-helm-repos
 	just enable-cnpg-default-namespace
 	just install-kube-prometheus-stack
+	just install-tempo
 	just install-traefik
 	just install-crd
 	just install-cnpg
--- a/tembo-operator/justfile
+++ b/tembo-operator/justfile
@@ -75,7 +75,6 @@ start-kind:
 	just update-helm-repos
 	just enable-cnpg-default-namespace
 	just install-kube-prometheus-stack
-	just install-tempo
 	just install-traefik
 	just install-crd
 	just install-cnpg
--- a/tembo-operator/src/controller.rs
+++ b/tembo-operator/src/controller.rs
@@ -3,8 +3,10 @@ use futures::stream::StreamExt;
 
 use crate::{
     apis::coredb_types::{CoreDB, CoreDBStatus},
-    cloudnativepg::backups::Backup,
-    cloudnativepg::cnpg::{cnpg_cluster_from_cdb, reconcile_cnpg, reconcile_cnpg_scheduled_backup},
+    cloudnativepg::{
+        backups::Backup,
+        cnpg::{cnpg_cluster_from_cdb, reconcile_cnpg, reconcile_cnpg_scheduled_backup},
+    },
     config::Config,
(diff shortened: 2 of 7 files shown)


tembo-operator/justfile
```diff
@@ -75,7 +75,6 @@ start-kind:
 	just update-helm-repos
 	just enable-cnpg-default-namespace
 	just install-kube-prometheus-stack
-	just install-tempo
```
ChuckHend, line 78: im not sure what this was doing, but just want to confirm that its meant to be deleted?

nhudson: Ah nope. I'll revert that

tembo-operator/src/controller.rs
```diff
@@ -598,6 +553,41 @@ impl CoreDB {
...
+            .metadata
+            .namespace
+            .clone()
+            .expect("CoreDB should have a namespace");
+        let backup: Api<Backup> = Api::namespaced(client, &namespace);
+        let lp = ListParams::default();
+        let backup_list = backup.list(&lp).await.map_err(|e| {
+            error!("Error getting backups: {:?}", e);
+            Action::requeue(Duration::from_secs(300))
+        })?;
```
sjmiller609, line 587: I think this part needs to somehow select the backups that are applicable for the current DB, since we cannot assume it's always one DB per namespace

tembo-operator/src/controller.rs
```diff
@@ -189,8 +184,6 @@ impl CoreDB {
         }
 
         debug!("Reconciling secret");
-        let span = span!(Level::INFO, "reconcile_secret");
```
nhudson, line 192: I'm also removing unneeded telemetry spans.

sjmiller609 (approved):
Very clean. One important consideration to handle, then looks good.

tembo-operator/src/controller.rs
nhudson: Sure.  That is easily done with adding a label filter.  I will make that change now.

sjmiller609 (approved):

tembo-operator/src/controller.rs
```diff
@@ -308,6 +283,7 @@ impl CoreDB {
                 trunk_installs: self.status.clone().and_then(|f| f.trunk_installs),
                 resources: Some(self.spec.resources.clone()),
                 runtime_config: self.spec.runtime_config.clone(),
+                first_recoverability_time: self.status.clone().and_then(|f| f.first_recoverability_time),
```
vrmiguel, line 286: Silly Rust nit: avoiding cloning `self.status`
```suggestion
                first_recoverability_time: self.status.as_ref().and_then(|f| f.first_recoverability_time),
```

nhudson: If we change that, then shouldn't the rest of these be updated as well?  Specifically line 283?

PR #270, opened 2023-09-21
set `restartedAt` and `status.running` correctly

vrmiguel:


conductor/src/lib.rs
```diff
@@ -429,6 +435,29 @@ pub async fn restart_cnpg(
     Ok(())
 }
 
+async fn patch_merge_cdb_status(
```
sjmiller609, line 438: this function already exists in controller.rs and is called
```
patch_cdb_status_merge
```

tembo-operator/src/cloudnativepg/cnpg.rs
```diff
@@ -537,6 +537,15 @@ pub async fn reconcile_cnpg(cdb: &CoreDB, ctx: Arc<Context>) -> Result<(), Actio
...
         .clone()
         .expect("CNPG Cluster should always have a name");
+
+    if let Some(restarted_at) = cdb.annotations().get(RESTARTED_AT) {
+        // Forward the `restartedAt` annotation from CoreDB over to the CNPG cluster
+        let mut cluster_annotations = cluster.metadata.annotations.unwrap_or_default();
+        cluster_annotations.insert(RESTARTED_AT.into(), restarted_at.to_owned());
+
+        cluster.metadata.annotations = Some(cluster_annotations);
+    }
```
sjmiller609, line 547: Here, I think we check if we are updating it to a different value, if so we set running to false on coredb

tembo-operator/src/extensions/database_queries.rs
```diff
@@ -136,6 +137,49 @@ pub async fn list_extensions(cdb: &CoreDB, ctx: Arc<Context>, database: &str) ->
...
+/// Returns Ok if the given database is running (i.e. not restarting)
+pub async fn is_not_restarting(cdb: &CoreDB, ctx: Arc<Context>, database: &str) -> Result<(), Action> {
+    let Some(restarted_at) = cdb.annotations().get(RESTARTED_AT) else {
+        // No restartedAt annotation, so we're not restarting
+        return Ok(());
+    };
+
+    let restarted_requested_at: DateTime<Utc> = DateTime::parse_from_rfc3339(restarted_at)
+        .map_err(|err| {
+            tracing::error!("Failed to deserialize DateTime from `restartedAt`: {err}");
```
sjmiller609, line 149: Can you please include cdb name in the log

tembo-operator/src/extensions/database_queries.rs
```diff
@@ -136,6 +137,49 @@ pub async fn list_extensions(cdb: &CoreDB, ctx: Arc<Context>, database: &str) ->
...
+
+    let pg_postmaster_start_time = cdb
+        .psql(
+            "select pg_postmaster_start_time();".to_owned(),
+            database.to_owned(),
+            ctx,
+        )
+        .await?
+        .stdout
+        .unwrap();
```
sjmiller609, line 163: Please handle stdout as None with unexpected error, long requeue

sjmiller609 (changes requested):
Nicely done!! A couple of small changes requested.

ianstanton (commented):
@vrmiguel Nicely done :tada: Can you add a test for this behavior in `tests/integration_tests.rs`?

tembo-operator/src/cloudnativepg/cnpg.rs
vrmiguel: Set running to false on `CoreDB` as in `cdb.running = false` ?

vrmiguel: As in

```rust
    if let Some(restarted_at) = cdb.annotations().get(RESTARTED_AT) {
        match existing_restarted_at_annotation {
            Some(timestamp) if timestamp == restarted_at => {}
            Some(_) | None => {
                // Forward the `restartedAt` annotation from CoreDB over to the CNPG cluster
                let mut cluster_annotations = cluster.metadata.annotations.unwrap_or_default();
                cluster_annotations.insert(RESTARTED_AT.into(), restarted_at.to_owned());
                drop(restarted_at);

                if let Some(status) = &mut cdb.status {
                    status.running = false;
                }

                cluster.metadata.annotations = Some(cluster_annotations);
            }
        }
    }

```

sjmiller609: In this block, I think either remember the annotation it was before the update or set a bool like restart_annotation_updated. Then down below in this function reconcile_cnpg, using that information we know if we have updated the Cluster with a new annotation. If we did, right after we update it ...
```
    let _o = cluster_api
        .patch(&name, &ps, &Patch::Apply(&cluster))
```

Then, call to patch cdb status like that function patch_merge_cdb_status. This will call kubernetes API to set cdb.status.running to false - we do this because we just detected that we are initiating a restart

sjmiller609 (approved):
Add a test or add to an existing test, like Ian has suggested, then I think it's good to go!

sjmiller609 (approved):

ianstanton (approved):
:tada: Nicely done!

PR #271, opened 2023-09-22
Add all config values to status

ianstanton:
This PR adds all current Postgres configuration values to `status.runtime_config`.

It also proposes a new directory structure `pkg/utils/`. This is a typical structure seen in Kubernetes Operator projects ([CNPG example](https://github.com/cloudnative-pg/cloudnative-pg/tree/main/pkg/utils)). A couple of reasons for this change:
- `extensions/database_queries.rs` is no longer specific to extensions, and will be used for general queries we execute across the codebase (examples in this PR and @vrmiguel's [recent PR](https://github.com/tembo-io/tembo-stacks/blob/d78016fc15192b9110019085eb103cdb9689f818/tembo-operator/src/extensions/database_queries.rs#L140-L181))
- We can move existing + future files with generic functions to this directory in an effort to de-clutter `src/`

Note:
- This PR will depend on some changes to control plane logic
- This dramatically increases the size of the resource's `status` section. It's hard to read without `| less`


tembo-operator/tests/integration_tests.rs
```diff
@@ -2599,4 +2600,154 @@ mod test {
...
+        let lp =
+            ListParams::default().labels(format!("app=postgres-exporter,coredb.io/name={}", name).as_str());
+        let exporter_pods = pods.list(&lp).await.expect("could not get pods");
+        let exporter_pod_name = exporter_pods.items[0].metadata.name.as_ref().unwrap();
+        println!("Exporter pod name: {}", &exporter_pod_name);
+
+        pod_ready_and_running(pods.clone(), exporter_pod_name.clone()).await;
+
+        // TODO(ianstanton) wait for status.runtime_config to be populated & updated with correct configs
+        sleep(Duration::from_secs(60));
```
ianstanton, line 2715: Placeholder while I investigate why it takes so long for the appropriate config values to show up

nhudson: yeah we need to avoid sleeping in tests, they end up causing more issues than it solves.  Hopefully you can figure that out.

ianstanton: :sleeping_bed:

tembo-operator/tests/integration_tests.rs
```diff
@@ -2599,4 +2600,154 @@ mod test {
...
+        let mut found_configs = false;
+        let expected_config = ConfigValue::Multiple(BTreeSet::from_iter(vec![
+            "pg_stat_statements".to_string(),
+            "pg_partman_bgw".to_string(),
+        ]));
+        for config in coredb_resource.status.unwrap().runtime_config.unwrap() {
+            if config.name == "shared_preload_libraries" {
+                println!("Found shared_preload_libraries config: {:?}", config);
+                found_configs = true;
+                assert_eq!(config.value, expected_config);
```
ianstanton, line 2727: Need to add some more assertions here :)

ChuckHend (approved):

tembo-operator/src/controller.rs
```diff
@@ -642,6 +648,15 @@ pub async fn get_current_coredb_resource(cdb: &CoreDB, ctx: Arc<Context>) -> Res
     Ok(coredb.clone())
 }
 
+// Get current config values
+pub async fn get_current_config_values(cdb: &CoreDB, ctx: Arc<Context>) -> Result<Vec<PgConfig>, Action> {
+    let cfg = list_config_params(cdb, ctx.clone()).await.map_err(|e| {
```
sjmiller609, line 653: since this is returning Action just use `?` I think

sjmiller609: this is important because it would be normal to fail the psql command under the condition where the pod is restarting, and in the case of a normal error, we requeue short instead of long.

tembo-operator/src/pkg/utils/database_queries.rs
```diff
@@ -136,6 +143,15 @@ pub async fn list_extensions(cdb: &CoreDB, ctx: Arc<Context>, database: &str) ->
     Ok(parse_extensions(&result_string))
 }
 
+/// List all configuration parameters
+pub async fn list_config_params(cdb: &CoreDB, ctx: Arc<Context>) -> Result<Vec<PgConfig>, Action> {
+    let psql_out = cdb
+        .psql("SHOW ALL;".to_owned(), "postgres".to_owned(), ctx)
+        .await?;
+    let result_string = psql_out.stdout.unwrap();
```
sjmiller609, line 151: Please handle the case of unwrap failing with long requeue and error log including coredb name

sjmiller609 (approved):
One feedback, looks great!

tembo-operator/src/pkg/utils/database_queries.rs
vrmiguel: Example from my PR

```rust
    let pg_postmaster_start_time = cdb
        .psql(
            "select pg_postmaster_start_time();".to_owned(),
            database.to_owned(),
            ctx,
        )
        .await?
        .stdout
        .ok_or_else(|| {
            tracing::error!("{cdb_name}: select pg_postmaster_start_time() had no stdout");

            Action::requeue(Duration::from_secs(300))
        })?;
   ```

**Comment by nhudson** [comment]
While I don't really have to much of a problem with it, the `pkg/utils` folder nesting is typical for a Golang based project.  Importing packages in Go is much cleaner than using the libraries in Rust.  It adds complexity to the Rust project, but overall works as long as you have all your `mod.rs` files in place pointing to the correct place.

**Review by nhudson — APPROVED** [review]

**Comment by ianstanton** [comment]
> While I don't really have to much of a problem with it, the `pkg/utils` folder nesting is typical for a Golang based project. Importing packages in Go is much cleaner than using the libraries in Rust. It adds complexity to the Rust project, but overall works as long as you have all your `mod.rs` files in place pointing to the correct place. I just hate seeing unnecessary files created and tracked just to keep a nested file structure. Either way as long as it works and we can all follow it, I am good with it.

Yeah, that is a good point. Let me see if there is a better solution for this. Maybe just moving the `database_queries.rs` file to `src`.

**Comment by ianstanton** [comment]
Something up with tests after resolving merge conflicts with main. Checking.

**Comment by ianstanton** [comment]
Back to fix things

**Comment by ianstanton** [comment]
Seems like there is an issue with this function https://github.com/tembo-io/tembo-stacks/pull/271/files#diff-a5d47cd69bbb4e60371bcfa07649e41efca99807609c58923c3e78ab93c442d9R147-R163. Reverting to draft while we fix

**Comment by ianstanton** [comment]
Legit failure in most recent commit - no stack overflow failure after adding --release flag

File: tembo-operator/src/app_service/manager.rs [hunk 5]
```diff
@@ -29,7 +29,6 @@ struct AppServiceResources {
     service: Option<Service>,
 }
 
-
```
ianstanton, line 32: From fmt

ianstanton:
Looks like we're in a good spot. Going to run the integration tests a few times to rule out any flaky tests.

PR #274, opened 2023-09-25
+Kube Service for appServices

ChuckHend:
Deploy a Kubernetes Service when appService has port mappings.


tembo-operator/src/app_service/manager.rs
```diff
@@ -165,6 +233,19 @@ async fn get_appservice_deployments(client: &Client, namespace: &str) -> Result<
         .collect())
 }
 
+// gets all names of AppService Services in the namespace that have the label "component=AppService"
```
sjmiller609, line 236: Probably should search in a way that does not collect the app services of other coredbs in the same namespace. For example adding a label to indicate which coredb it's connected to and add that to the label selector

tembo-operator/src/app_service/manager.rs
```diff
@@ -234,34 +406,15 @@ pub async fn reconcile_app_services(cdb: &CoreDB, ctx: Arc<Context>) -> Result<(
...
-                    ns, ad.name, e
-                );
-            }
-        }
-    }
 
-    if has_errors {
+    let apply_errored = apply_resources(resources, &client, &ns).await;
+
+    if has_errors || apply_errored {
```
sjmiller609, line 417: Probably in a future change, we would also want to check for readiness of pods associated to the deployments.

For example if a stack has a default app service included, this would return, and so the whole reconcile process finishes, before the app service is ready - that would end up like goes from "submitted" to "good" but maybe an app service is still loading.

If checking for app services ready, that would look like after we do this block to apply them, then we check if they are ready. If they are not ready, then short requeue.

sjmiller609 (approved):

tembo-operator/src/app_service/manager.rs
ChuckHend: Agreed. I have a whole issue in our backlog for more advanced error handling and reporting.

tembo-operator/src/app_service/manager.rs
ChuckHend: This also applies to the existing Deployments then right? Up until now I have been assuming 1 coredb instance per namespace, since thats how it runs in tembo-cloud. Do we want to change that so we'll have multiple coredb in a single namespace?

tembo-operator/tests/integration_tests.rs
```diff
@@ -2833,21 +2862,22 @@ mod test {
...
+        let service_items: Vec<Service> = list_resources(client.clone(), cdb_name, &namespace, 0)
+            .await
+            .unwrap();
+        assert!(service_items.is_empty());
+        // should be no Services
+        let service_items: Vec<Service> = list_resources(client.clone(), cdb_name, &namespace, 0)
+            .await
+            .unwrap();
+        // One appService Services
+        assert!(service_items.is_empty());
```
ianstanton, line 2879:
```suggestion
```
Looked like duplicate

tembo-operator/tests/integration_tests.rs
```diff
@@ -315,6 +316,50 @@ mod test {
...
     }
 
+    use controller::errors;
+    use k8s_openapi::NamespaceResourceScope;
+    use serde::de::DeserializeOwned;
+
+    // helper function retrieve all instances of a resource in namespace
+    // used repeatedly in appService tests
+    // handles retries
+    async fn list_resources<R>(
```
ianstanton, line 326: Super useful :smile:

tembo-operator/src/app_service/manager.rs
```diff
@@ -226,6 +383,21 @@ pub async fn reconcile_app_services(cdb: &CoreDB, ctx: Arc<Context>) -> Result<(
         }
     }
 
+    // reap any AppService  that are no longer desired
+    if let Some(to_delete) = appservice_to_delete(desired_services, actual_services) {
```
ianstanton, line 387: Just a thought (not sure if we want to do this yet), but if we introduced `AppService` as a new CRD like `CoreDB` it might improve handling create / delete of deployments and services. For example, instead of looking for deployments and services to delete, we'd just look for `AppServices` to delete and `OwnerReferences` would take care of cleaning up relevant deployments + services owned by that `AppService`

ianstanton (approved):

tembo-operator/src/app_service/manager.rs
ChuckHend: @sjmiller609 
I updated the GET Deployments and Services to include `format!("component={},coredb.io/name={}", COMPONENT_NAME, coredb_name);` in the ListParams.

PR #275, opened 2023-09-27
[tembo-operator] Add instance restore capabilities into operator

nhudson:
Restore work will be split up into a few different PR's.  This change adds basic restore capabilities into the controller.

* Add support to run Minio locally to test backup and restores with. ([TEM-1834](https://linear.app/tembo/issue/TEM-1834/add-ability-to-add-tests-for-backups-and-restores))
* Add integration test to run an instance, add some data, back it up then restore it and check if the data is on the new restored instance.
* Adds support into the operator for non-AWS S3 compatible object storage (Minio, DigitalOcean, etc)
* Basic support to restore an instance from a backup in the `CoreDBSpec`.

** Reconciling a set of extensions into the restored instance and Conductor support will come in a later PR's

fixes: [TEM-1379](https://linear.app/tembo/issue/TEM-1379/bootstrap-from-configuration-in-operator)

### DIFF
--- a/charts/tembo-operator/templates/crd.yaml
+++ b/charts/tembo-operator/templates/crd.yaml
@@ -125,22 +125,77 @@ spec:
               backup:
                 default:
                   destinationPath: s3://
-                  encryption: AES256
+                  encryption: null
                   retentionPolicy: '30'
                   schedule: 0 0 * * *
+                  endpointURL: null
+                  s3Credentials: null
                 properties:
                   destinationPath:
                     default: s3://
                     nullable: true
                     type: string
                   encryption:
-                    default: AES256
+                    nullable: true
+                    type: string
+                  endpointURL:
                     nullable: true
                     type: string
                   retentionPolicy:
                     default: '30'
                     nullable: true
                     type: string
+                  s3Credentials:
+                    nullable: true
+                    properties:
+                      accessKeyId:
+                        nullable: true
+                        properties:
+                          key:
+                            type: string
+                          name:
+                            type: string
+                        required:
+                        - key
+                        - name
+                        type: object
+                      inheritFromIAMRole:
+                        nullable: true
+                        type: boolean
+                      region:
+                        nullable: true
+                        properties:
+                          key:
+                            type: string
+                          name:
+                            type: string
+                        required:
+                        - key
+                        - name
+                        type: object
+                      secretAccessKey:
+                        nullable: true
+                        properties:
+                          key:
+                            type: string
+                          name:
+                            type: string
+                        required:
+                        - key
+                        - name
+                        type: object
+                      sessionToken:
+                        nullable: true
+                        properties:
+                          key:
+                            type: string
+                          name:
+                            type: string
+                        required:
+                        - key
+                        - name
+                        type: object
+                    type: object
                   schedule:
                     default: 0 0 * * *
                     nullable: true
@@ -254,6 +309,71 @@ spec:
                       type: string
                     description: 'Requests describes the minimum amount of compute resources required. If Requests is omitted for a container, it defaults to Limits if that is explicitly specified, otherwise to an implementation-defined value. More info: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/'
                     type: object
+                type: object
+              restore:
+                nullable: true
+                properties:
+                  endpointURL:
+                    nullable: true
+                    type: string
+                  recoveryTargetTime:
+                    nullable: true
+                    type: string
+                  s3Credentials:
+                    nullable: true
+                    properties:
+                      accessKeyId:
+                        nullable: true
+                        properties:
+                          key:
+                            type: string
+                          name:
+                            type: string
+                        required:
+                        - key
+                        - name
+                        type: object
+                      inheritFromIAMRole:
+                        nullable: true
+                        type: boolean
+                      region:
+                        nullable: true
+                        properties:
+                          key:
+                            type: string
+                          name:
+                            type: string
+                        required:
+                        - key
+                        - name
+                        type: object
+                      secretAccessKey:
+                        nullable: true
+                        properties:
+                          key:
+                            type: string
+                          name:
+                            type: string
+                        required:
+                        - key
+                        - name
+                        type: object
+                      sessionToken:
+                        nullable: true
+                        properties:
+                          key:
+                            type: string
+                          name:
+                            type: string
+                        required:
+                        - key
+                        - name
+                        type: object
+                    type: object
+                  serverName:
+                    type: string
+                required:
+                - serverName
                 type: object
               runtime_config:
                 items:
--- a/charts/tembo-operator/templates/crd.yaml
+++ b/charts/tembo-operator/templates/crd.yaml
@@ -125,17 +125,19 @@ spec:
               backup:
                 default:
                   destinationPath: s3://
-                  encryption: null
+                  encryption: AES256
                   retentionPolicy: '30'
                   schedule: 0 0 * * *
                   endpointURL: null
-                  s3Credentials: null
+                  s3Credentials:
+                    inheritFromIAMRole: true
                 properties:
                   destinationPath:
                     default: s3://
                     nullable: true
                     type: string
                   encryption:
+                    default: AES256
                     nullable: true
                     type: string
                   endpointURL:
@@ -146,6 +148,8 @@ spec:
                     nullable: true
                     type: string
                   s3Credentials:
+                    default:
+                      inheritFromIAMRole: true
                     nullable: true
                     properties:
                       accessKeyId:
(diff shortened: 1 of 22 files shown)


conductor/src/main.rs
```diff
@@ -242,6 +242,7 @@ async fn run(metrics: CustomMetrics) -> Result<(), Box<dyn std::error::Error>> {
                     encryption: Some(String::from("AES256")),
                     retentionPolicy: Some(String::from("30")),
                     schedule: Some(generate_rand_schedule().await),
+                    ..Default::default()
```
nhudson, line 245: This is to just satisfy the requirements of the struct.  Conductor as of now would not generate a backup configuration to a non AWS S3 bucket.

tembo-operator/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "controller"
 description = "Tembo Operator for Postgres"
-version = "0.4.3"
+version = "0.5.0"
```
nhudson, line 4: This feels like a big enough change to force the version to a `0.5.0`.  If anyone thinks this doesn't need to happen, please let me know.

tembo-operator/justfile
```diff
@@ -42,6 +42,24 @@ install-tempo:
 		tempo grafana/tempo \
 	  --namespace monitoring
 
+install-minio:
```
nhudson, line 45: Install minio into the kind cluster to simulate backup and restore.

tembo-operator/src/apis/coredb_types.rs
```diff
@@ -34,17 +30,74 @@ pub struct ServiceAccountTemplate {
     pub metadata: Option<ObjectMeta>,
 }
 
-#[derive(Deserialize, Serialize, Clone, Debug, JsonSchema, Default)]
+#[derive(Serialize, Deserialize, Clone, Debug, Default, JsonSchema)]
+pub struct S3Credentials {
```
nhudson, line 34: With this technically out  operator can backup/restore from any s3 compatible object storage now.

tembo-operator/src/apis/coredb_types.rs
```diff
@@ -34,17 +30,74 @@ pub struct ServiceAccountTemplate {
...
     #[serde(default = "defaults::default_destination_path")]
     pub destinationPath: Option<String>,
-    #[serde(default = "defaults::default_encryption")]
     pub encryption: Option<String>,
     #[serde(default = "defaults::default_retention_policy")]
     pub retentionPolicy: Option<String>,
     #[serde(default = "defaults::default_backup_schedule")]
     pub schedule: Option<String>,
+    #[serde(default, rename = "endpointURL")]
+    pub endpoint_url: Option<String>,
```
nhudson, line 86: These new enums are needed to work with non AWS S3 object storage (eg Minio)

tembo-operator/src/controller.rs
```diff
@@ -799,7 +799,7 @@ mod test {
 
         let expected_time = NaiveDate::from_ymd_opt(2023, 9, 19)
             .and_then(|date| date.and_hms_opt(23, 14, 0))
-            .map(|naive_dt| DateTime::<Utc>::from_utc(naive_dt, Utc));
```
nhudson, line 802: `from_utc` is deprecated, so I updated it to the new method `from_naive_utc_and_offset`

tembo-operator/src/extensions/install.rs
```diff
@@ -51,8 +51,6 @@ fn merge_and_deduplicate_pods(non_fenced_pods: Vec<Pod>, fenced_names: Option<Ve
 // Collect any fenced pods and add them to the list of pods to install extensions into
 #[instrument(skip(ctx, cdb) fields(trace_id))]
 async fn all_fenced_and_non_fenced_pods(cdb: &CoreDB, ctx: Arc<Context>) -> Result<Vec<Pod>, Action> {
-    let span = span!(Level::DEBUG, "all_fenced_and_non_fenced_pods");
```
nhudson, line 54: Removing unneeded spans

tembo-operator/Cargo.toml
nhudson: Depends on what goes in first this might need to bump to `0.6.0`

tembo-operator/justfile
```diff
@@ -74,9 +94,11 @@ start-kind:
 	kind create cluster --image=kindest/node:v{{KUBE_VERSION}} --config testdata/kind-config.yaml
 	just update-helm-repos
 	just enable-cnpg-default-namespace
+	just install-reflector
```
nhudson, line 97: Installing reflector, to copy the minio creds to all namespaces.

tembo-operator/src/apis/coredb_types.rs
```diff
@@ -34,17 +30,74 @@ pub struct ServiceAccountTemplate {
...
+    pub key: String,
+    pub name: String,
+}
+
+#[derive(Deserialize, Serialize, Clone, Debug, Default, JsonSchema)]
 #[allow(non_snake_case)]
 pub struct Backup {
     #[serde(default = "defaults::default_destination_path")]
     pub destinationPath: Option<String>,
-    #[serde(default = "defaults::default_encryption")]
```
nhudson, line 42: Removing the default encryption call here since by default it needs to be `None` now.  We set this in conductor.

tembo-operator/src/cloudnativepg/cnpg.rs
```diff
@@ -98,55 +134,163 @@ pub fn cnpg_backup_configuration(
...
+        match result {
+            Ok(parsed_time) => Ok(Some(parsed_time)),
+            Err(err) => Err(ValueError::ChronoParseError(err)),
+        }
+    } else {
+        Ok(None)
+    }
+}
+
+#[instrument(skip(cdb))]
```
nhudson, line 244: This entire function was pretty time consuming to work with.  In the end we can accept all of these time formats, but we convert it to a single format `%Y-%m-%d %H:%M:%S.6T+TZ`

tembo-operator/src/cloudnativepg/cnpg.rs
```diff
@@ -1077,6 +1245,111 @@ pub async fn unfence_pod(cdb: &CoreDB, ctx: Arc<Context>, pod_name: &str) -> Res
     }
 }
 
+// generate_restore_destination_path function will generate the restore destination path from the backup
+// object and return a string
+fn generate_restore_destination_path(path: &str) -> String {
+    let mut parts: Vec<&str> = path.split('/').collect();
+    parts.pop();
+    parts.join("/")
+}
```
nhudson, line 1254: This function is needed to generate the restore destination bucket path for the s3 bucket.  We take the input from the backup destination bucket path.  If we wanted to allow from one different bucket path to another, then there would be more work to be done.  The assumption here is that the s3 path will be the same for backup and restore.

charts/tembo-operator/templates/crd.yaml
```diff
@@ -129,22 +129,77 @@ spec:
               backup:
                 default:
                   destinationPath: s3://
-                  encryption: AES256
+                  encryption: null
                   retentionPolicy: '30'
                   schedule: 0 0 * * *
+                  endpointURL: null
```
sjmiller609, line 135: We will need endpoint URL for cost control because otherwise we are using NAT gateway and pay egress fees on backup (I think, not 100% sure).

tembo-operator/justfile
sjmiller609: ok so this is why we need the S3 credentials added?

tembo-operator/src/apis/coredb_types.rs
sjmiller609: why not default to AES256, and configure it the other way if required for minio?

tembo-operator/tests/integration_tests.rs
```diff
@@ -3088,4 +3124,280 @@ mod test {
...
+        // Generate basic CoreDB resource to start with
+        let coredb_json = serde_json::json!({
+            "apiVersion": API_VERSION,
+            "kind": kind,
+            "metadata": {
+                "name": name
+            },
+            "spec": {
+            "replicas": replicas,
+                "backup": {
```
sjmiller609, line 3167: I mean maybe here in the test we can set encryption to None if we must disable it for minio?

sjmiller609 (approved):
Looks clean:
- I was hesitant for the s3 creds additions to CRD, but seems justified to allow for local testing with minio or other object storages which are typically all S3 compatible APIs
- I think should keep default encryption if possible to unset the default in the test for minio compatibility.

charts/tembo-operator/templates/crd.yaml
nhudson: yeah we can set the s3 endpoint when we get that setup.  This can be something that can be figured out in conductor and set in `CoreDB` depending on the deployment region.

tembo-operator/tests/integration_tests.rs
nhudson: encryption is now set back to AES256 by default.

sjmiller609 (approved):

PR #282, opened 2023-09-28
conductor: check `status.running` after restarting

vrmiguel:


nhudson (approved):

conductor/src/main.rs
```diff
@@ -427,25 +425,43 @@ async fn run(metrics: CustomMetrics) -> Result<(), Box<dyn std::error::Error>> {
...
-                    metrics
-                        .conductor_errors
-                        .add(&opentelemetry::Context::current(), 1, &[]);
-                    continue;
-                }
-                let current_resource = result?;
-                let resource_json = serde_json::to_string(&current_resource).unwrap();
-                debug!("dbname: {}, current: {:?}", &namespace, resource_json);
+
+                let result = get_coredb_error_without_status(client.clone(), &namespace).await;
```
vrmiguel, line 429: Is there any chance `status` would be `None` here after a restart request? cc @sjmiller609

ianstanton (approved):

conductor/src/main.rs
ianstanton: I don't think so

PR #286, opened 2023-09-29
add pg conn secret to appService containers

ChuckHend:
- adds example using Postgres
- changes env var type in CRD so that we can map one of the Tembo injected env vars to user provided env var
For example - map the existing env var for the "ReadWriteConnection" to the user provided "PGRST_DB_URI" var.
```
  "env": [
      {
          "name": "PGRST_DB_URI",
          "valueFromEnv": "ReadWriteConnection"
      }
 ```
- adds a test to send http request to an `appService`


sjmiller609 (approved):
Maybe valueFromSecret?

Also maybe we should make it match names here https://api.data-1.use1.tembo.io/swagger-ui/#/secrets/get_secret_names

Also secrets have multiple key value pairs, sometimes an app might want the username and password separately for example, but that’s inside the same secret.

ChuckHend:
> Also secrets have multiple key value pairs, sometimes an app might want the username and password separately for example, but that’s inside the same secret.

Correct, and we are adding a few new key value pairs in this PR: https://github.com/tembo-io/tembo-stacks/blob/65a12442b549c9be6d6d5a5146b17e914c07a950/tembo-operator/src/secret.rs#L118-L120 . This PR selects a subset of those secret values and adds them as environment variables in the appService container.

nhudson (approved):

PR #287, opened 2023-09-29
Add namespace annotations

sjmiller609:


conductor/src/lib.rs
```diff
@@ -190,6 +195,10 @@ pub async fn create_namespace(client: Client, name: &str) -> Result<(), Conducto
         "kind": "Namespace",
         "metadata": {
             "name": format!("{name}"),
+            "annotations": {
+                "tembo.io/instance_id": instance_id,
+                "tembo.io/organization_id": organization_id
+            },
```
sjmiller609, line 201: This allows for mapping between org ID and namespace

ianstanton (approved):

nhudson (approved):

PR #290, opened 2023-09-29
conductor: log the restartedAt annotation being set

vrmiguel:


sjmiller609 (approved):

conductor/src/lib.rs
```diff
@@ -427,6 +427,8 @@ pub async fn restart_cnpg(
         }
     });
 
+    info!("Applied `restartedAt == {restart}` to the CoreDB resource. Setting `status.running = false`.");
```
sjmiller609, line 430:
```suggestion
    info!("Applying `restartedAt == {restart}` to the CoreDB resource. Setting `status.running = false`.");
```

PR #296, opened 2023-10-02
Use labels instead of annotations for namespace

sjmiller609:
to look up namespaces we use labels instead of annotations https://docs.rs/kube/latest/kube/api/struct.ListParams.html


sjmiller609:
while making the change in data API, I realized this should have been labels to perform a filter on all namespaces with this.

nhudson (approved):

PR #299, opened 2023-10-02
Increase the stack size per thread

sjmiller609:


nhudson (approved):
fix linting, then you should be fine.

vrmiguel (approved):

tembo-operator/src/main.rs
```diff
@@ -46,3 +45,11 @@ async fn main() -> anyhow::Result<()> {
     tokio::join!(controller, server.run()).1?;
     Ok(())
 }
+
+fn main() -> anyhow::Result<()> {
+    let rt = tokio::runtime::Builder::new_multi_thread()
+        .enable_all()
+        .thread_stack_size(4 * 1024 * 1024)
```
vrmiguel, line 52: Explanation: the default is 2MiB, we're now doubling it

PR #301, opened 2023-10-02
operator: compare CNPG `restartedAt` using the current spec

vrmiguel:
Fix `restartedAt` comparison: was previously using the value from the new cluster spec being created in `reconcile_cnpg`.

Based off of previous work done with @sjmiller609 on #300.

### DIFF
--- a/tembo-operator/src/cloudnativepg/cnpg.rs
+++ b/tembo-operator/src/cloudnativepg/cnpg.rs
@@ -741,6 +741,49 @@ async fn pods_to_fence(cdb: &CoreDB, ctx: Arc<Context>) -> Result<Vec<String>, A
     }
 }
 
+fn update_restarted_at(cdb: &CoreDB, maybe_cluster: Option<&Cluster>, new_spec: &mut Cluster) -> bool {
+    let Some(cdb_restarted_at) = cdb.annotations().get(RESTARTED_AT) else {
+        // Avoid interacting with k8s if CoreDB does not have
+        // a restartedAt tag
+        return false;
+    };
+
+    let Some(current_cluster) = maybe_cluster else {
+        return true;
+    };
+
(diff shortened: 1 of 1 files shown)


tembo-operator/src/cloudnativepg/cnpg.rs
```diff
@@ -741,6 +741,49 @@ async fn pods_to_fence(cdb: &CoreDB, ctx: Arc<Context>) -> Result<Vec<String>, A
...
+        });
+
+        let name = new_spec.metadata.name.as_deref().unwrap_or("unknown");
+        info!("restartAt changed for cluster {name}, setting to {cdb_restarted_at}.");
+    }
+
+    restart_annotation_updated
+}
+
+fn did_restarted_at_change(cdb: &CoreDB, cluster: &Cluster) -> bool {
```
vrmiguel, line 770: Separated as a unit-testable function (tests yet to be done)

tembo-operator/src/cloudnativepg/cnpg.rs
```diff
@@ -763,34 +806,10 @@ pub async fn reconcile_cnpg(cdb: &CoreDB, ctx: Arc<Context>) -> Result<(), Actio
...
-            }
-        }
-    } else {
-        false
-    };
-
     let cluster_api: Api<Cluster> = Api::namespaced(ctx.client.clone(), namespace.as_str());
+    let maybe_cluster = cluster_api.get(&name).await;
+
+    let restart_annotation_updated = update_restarted_at(cdb, maybe_cluster.as_ref().ok(), &mut cluster);
```
vrmiguel, line 812: This was done to reutilize the current `Cluster` config that was previously fetched below

This maintains the previous behavior upon `Api::get` failure while allowing us to use the Cluster if it was found

sjmiller609 (approved):

PR #305, opened 2023-10-03
force minio to use less memory

nhudson:


tembo-operator/testdata/minio.yaml
```diff
@@ -2,7 +2,11 @@ mode: standalone
 persistence:
   enabled: true
   size: 10Gi
-resources: {}
+resources:
+  requests:
+    memory: 100Mi
+  limits:
+    memory: 2Gi
```
sjmiller609, line 9:
```suggestion
resources:
  requests:
    memory: ~
```
I found this works

PR #310, opened 2023-10-04
Enable pg_partman_bgw in shared_preload_libraries

sjmiller609:


tembo-operator/src/apis/coredb_types.rs
```diff
@@ -174,7 +175,10 @@ pub struct CoreDBSpec {
 impl CoreDBSpec {
     // extracts all postgres configurations
     // configs can be defined in several different places (from a stack, user override, from an extension installation, user overrides, etc)
-    pub fn get_pg_configs(&self, requires_load: Vec<String>) -> Result<Option<Vec<PgConfig>>, MergeError> {
+    pub fn get_pg_configs(
+        &self,
+        requires_load: BTreeMap<String, String>,
```
sjmiller609, line 180: throughout this change, instead of requires load a list of extension names, now it's a map of extension names to library-to-be-loaded name, [which are not always exact match.](https://tembo.io/blog/four-types-of-extensions/#load:~:text=Libraries%20are%20not%20always%20named%20exactly%20the%20same%20as%20the%20extension.)

PR #316, opened 2023-10-06
More specific security policy for app services

sjmiller609:


tembo-operator/src/app_service/manager.rs
```diff
@@ -182,9 +183,21 @@ fn generate_deployment(
         None => None,
     };
 
+    // https://tembo.io/docs/tembo-cloud/security/#tenant-isolation
```
sjmiller609, line 186: https://tembo.io/docs/tembo-cloud/security/#tenant-isolation

ChuckHend (approved):

PR #317, opened 2023-10-06
IP allow list

sjmiller609:


tembo-operator/src/ingress.rs
```diff
@@ -10,19 +10,37 @@ use kube::{
...
-use crate::{apis::coredb_types::CoreDB, errors::OperatorError, Context};
+use crate::{
+    apis::coredb_types::CoreDB,
+    errors::OperatorError,
+    traefik::middleware_tcp_crd::{MiddlewareTCP, MiddlewareTCPIpWhiteList, MiddlewareTCPSpec},
+    Context,
+};
 use tracing::{debug, error, info, warn};
 
+pub const VALID_IPV4_CIDR_BLOCK: &str = "^((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.){3}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)(/(3[0-2]|2[0-9]|1[0-9]|[0-9]))?$";
```
sjmiller609, line 24: We can import this in the control plane to use for input validation up front

tembo-operator/src/ingress.rs
```diff
@@ -337,3 +401,138 @@ pub async fn reconcile_postgres_ing_route_tcp(
 
     Ok(())
 }
+
+fn generate_ip_allow_list_middleware_tcp(cdb: &CoreDB) -> MiddlewareTCP {
```
sjmiller609, line 405: This function takes the approach if there are zero specified, it is default-open. It is more secure to default-closed, but I think it's better as an initial step to default-open, meaning if there is nothing specified or nothing that is specified is valid, then it's open.

ChuckHend (approved):

PR #324, opened 2023-10-09
Update Traefik Custom Resource Definitions

sjmiller609:
This PR updates the tests to use Traefik CRDs defined in the 3.0.0-beta.2 release of Traefik

https://github.com/traefik/traefik-helm-chart/issues/757#issuecomment-1753995542


sjmiller609:
I should also re-run kopium. I think there are basically fields that won't match, but we are not using those fields yet (until ip allow list)

ianstanton (approved):

ChuckHend (approved):

PR #330, opened 2023-10-12
Local dev suite

DarrenBaldwin07:


conductor/src/main.rs
```diff
@@ -53,6 +53,10 @@ async fn run(metrics: CustomMetrics) -> Result<(), ConductorError> {
         .unwrap_or_else(|_| "100".to_owned())
         .parse()
         .expect("error parsing MAX_READ_CT");
+    let is_cloud_formation: bool = env::var("IS_CLOUD_FORMATION")
+    .unwrap_or_else(|_| "100".to_owned())
+    .parse()
+    .expect("error parsing MAX_READ_CT");
```
ChuckHend, line 59: should this be a bool?

DarrenBaldwin07: Good catch - the `unwrap_or_else` should be set to `true`

sjmiller609 (approved):

PR #340, opened 2023-10-17
Add `pooler_host` to `ConnectionInfo` in conductor

ianstanton:
Add `pooler_host` field to `ConnectionInfo`. This information is passed up to the control-plane.

### DIFF
--- a/conductor/src/types.rs
+++ b/conductor/src/types.rs
@@ -54,7 +54,7 @@ pub struct OrgInstId {
 #[derive(Debug, Serialize, Deserialize)]
 pub struct ConnectionInfo {
     pub host: String,
-    pub pooler_host: String,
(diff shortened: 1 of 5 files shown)


conductor/src/types.rs
```diff
@@ -54,6 +54,7 @@ pub struct OrgInstId {
 #[derive(Debug, Serialize, Deserialize)]
 pub struct ConnectionInfo {
     pub host: String,
+    pub pooler_host: String,
```
ChuckHend, line 57: Will this be an Option, or there will always be a value even when Pooler not enabled?

ianstanton: It should be optional. I'll update

ChuckHend (approved):

sjmiller609 (approved):

PR #346, opened 2023-10-19
Add a common CA for postgres servers

sjmiller609:
so that users only need to add 1 CA to their trusted store.

Example of underlying CA

```
---
apiVersion: cert-manager.io/v1
kind: ClusterIssuer
metadata:
  name: selfsigning-postgres-ca-issuer
spec:
  selfSigned: {}
---
apiVersion: cert-manager.io/v1
kind: Certificate
metadata:
  name: postgres-selfsigned-ca
  namespace: cert-manager
spec:
  dnsNames:
  - localhost
  - local.tembo-development.com
  isCA: true
  issuerRef:
    group: cert-manager.io
    kind: ClusterIssuer
    name: selfsigning-postgres-ca-issuer
  privateKey:
    algorithm: ECDSA
    size: 256
  secretName: postgres-ca-secret
---
apiVersion: cert-manager.io/v1
kind: ClusterIssuer
metadata:
  name: postgres-server-issuer
spec:
  ca:
    secretName: postgres-ca-secret
```

In the above example, a CA is issued from a self-signed issuer. Then, another issuer is created to issue from that CA.

We extract the ca.crt from the centralized secret and sync it into each namespace, without copying the private key. Then, we issue a server and replication certificate for each server from the common CA. 

In a subsequent change, I will need to coordinate flagging this on and reconfiguring dataplane webserver to use new ca secret name at the same time.


tembo-operator/src/postgres_certificates.rs
```diff
@@ -0,0 +1,178 @@
...
+const POSTGRES_CA_SECRET_NAME: &str = "postgres-ca-secret";
+const POSTGRES_CA_SECRET_CERT_KEY_NAME: &str = "ca.crt";
+const POSTGRES_CERTIFICATE_ISSUER_NAME: &str = "postgres-server-issuer";
+
+pub async fn reconcile_certificates(
+    client: Client,
+    coredb_name: &str,
+    namespace: &str,
+) -> Result<(), Action> {
+    match std::env::var("USE_SHARED_CA") {
```
sjmiller609, line 24: It's flagged off by default, but it's turned on in the tests

ChuckHend (approved):

nhudson (approved):

PR #353, opened 2023-10-25
Correct issue with setting up CA and Secret for Connection Pooler

nhudson:
There is an issue with setting up the `-pooler` secret since we switched to using our own generated CA.  This PR is an attempt to add support to generate the certificate and secret the pooler requires to reconcile along with setting up the `cnpg_pooler_pgbouncer` role in the database.

fixes: [TEM-2039](https://linear.app/tembo/issue/TEM-2039/fix-pooler-mode-to-work-with-custom-certs)

### DIFF
--- a/.github/workflows/operator.yaml
+++ b/.github/workflows/operator.yaml
@@ -75,6 +75,7 @@ jobs:
         run: |
           set -xe
           export DATA_PLANE_BASEDOMAIN=localhost
+          export USE_SHARED_CA=1
           # Start the operator in the background
           cargo run > operator-output.txt 2>&1 &
           # Run the tests
--- a/tembo-operator/src/cloudnativepg/cnpg.rs
+++ b/tembo-operator/src/cloudnativepg/cnpg.rs
@@ -39,11 +39,8 @@ use crate::{
     config::Config,
     defaults::{default_image, default_llm_image},
     errors::ValueError,
-    extensions::database_queries::{
-        POOLER_CREATE_FUNCTION, POOLER_CREATE_ROLE, POOLER_GRANT_ON_FUNCTION, POOLER_GRANT_TO_APP,
-        POOLER_GRANT_TO_POSTGRES, POOLER_REVOKE_ON_FUNCTION,
-    },
     patch_cdb_status_merge,
+    psql::PsqlOutput,
     trunk::extensions_that_require_load,
     Context, RESTARTED_AT,
 };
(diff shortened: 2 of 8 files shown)


.github/workflows/operator.yaml
```diff
@@ -75,6 +75,7 @@ jobs:
         run: |
           set -xe
           export DATA_PLANE_BASEDOMAIN=localhost
+          export USE_SHARED_CA=1
```
nhudson, line 78: This will force the CI tests to use the new shared CA.  This was causing some false positives.

tembo-operator/src/cloudnativepg/cnpg.rs
```diff
@@ -976,14 +977,11 @@ pub async fn reconcile_cnpg(cdb: &CoreDB, ctx: Arc<Context>) -> Result<(), Actio
         return Err(Action::requeue(Duration::from_secs(10)));
     }
 
-    // Reconcile Pooler resource
-    reconcile_pooler(cdb, ctx.clone()).await?;
-
```
nhudson, line 981: Due to an order of operations I moved the reconcile loop for the pooler to the `controllers.rs` code file and it will reconcile after the postgres exporter is reconciled.

.github/workflows/operator.yaml
sjmiller609: cc @vitezdev this is why we had locally failing but passing in CI

sjmiller609 (approved):

nhudson:
https://linear.app/tembo/issue/TEM-2039/fix-pooler-mode-to-work-with-custom-certs

nhudson:
```
k get secrets my-coredb-ca1 -ojson | jq -r '.data["ca.crt"]' | base64 -d >ca.crt
❯ psql "postgres://postgres:$(kubectl get secrets -o json my-coredb-connection | jq -r '.data.password' | base64 --decode)@my-coredb-pooler.localhost:5432?sslmode=verify-full&sslrootcert=ca.crt"
psql (15.4, server 15.3)
SSL connection (protocol: TLSv1.3, cipher: TLS_AES_256_GCM_SHA384, compression: off)
Type "help" for help.

postgres=# \dx;
                   List of installed extensions
    Name    | Version |   Schema   |         Description          
------------+---------+------------+------------------------------
 pg_graphql | 1.2.0   | graphql    | pg_graphql: GraphQL support
 plpgsql    | 1.0     | pg_catalog | PL/pgSQL procedural language
(2 rows)

postgres=# 
```

PR #357, opened 2023-10-26
add default resource values for appService

ChuckHend:


charts/tembo-operator/templates/crd.yaml
```diff
@@ -152,8 +152,14 @@ spec:
                       - readiness
                       type: object
                     resources:
+                      default:
+                        limits:
+                          cpu: '2'
+                          memory: 2Gi
+                        requests:
+                          cpu: 500m
+                          memory: 512Mi
```
sjmiller609, line 161: Kinda big default, also I think default should have no oversubscription for memory.

```suggestion
                      default:
                        limits:
                          cpu: '1'
                          memory: 512Mi
                        requests:
                          cpu: 250m
                          memory: 512Mi
```

sjmiller609 (approved):

charts/tembo-operator/templates/crd.yaml
ChuckHend: good catch. let me regenerate the CRD

ChuckHend: 100m and 400m for CPU sound good?

sjmiller609 (approved):

charts/tembo-operator/templates/crd.yaml
sjmiller609: Yes it looks good

PR #369, opened 2023-11-03
Add heartbeat table to ensure PITR will work on databases with no transactions

nhudson:
We have seen where instances cannot be restored due to little to no transactions taking place, essentially an empty database.  The issue is that a date and time can be requested, but since there are no transactions during that time you want to restore to the restore will fail.

This adds a new schema (tembo) and a function that will write a timestamp to a table (heartbeat_timestamp).

Feel free to make suggestions on how things should be named.

fixes: [TEM-2132](https://linear.app/tembo/issue/TEM-2132/pitr-will-fail-on-instances-with-zero-transactions)

### DIFF
--- a/tembo-operator/src/heartbeat.rs
+++ b/tembo-operator/src/heartbeat.rs
@@ -37,11 +37,14 @@ BEGIN
         );';
     END IF;
 
+    -- Create index on latest_heartbeat column
+    EXECUTE 'CREATE INDEX idx_heartbeat ON tembo.heartbeat_table (latest_heartbeat);';
+
     -- Insert current UTC timestamp into heartbeat_table
     EXECUTE 'INSERT INTO tembo.heartbeat_table (latest_heartbeat)
         VALUES (CURRENT_TIMESTAMP AT TIME ZONE ''UTC'');';
 
-    -- Delete entries older than 30 days
+    -- Delete entries older than 7 days
     EXECUTE 'DELETE FROM tembo.heartbeat_table
         WHERE latest_heartbeat < (CURRENT_TIMESTAMP AT TIME ZONE ''UTC'' - INTERVAL ''7 days'');';
 
--- a/tembo-operator/src/heartbeat.rs
+++ b/tembo-operator/src/heartbeat.rs
@@ -1,0 +1,91 @@
+use crate::{apis::coredb_types::CoreDB, psql::PsqlOutput, Context};
+use kube::{runtime::controller::Action, ResourceExt};
+use std::sync::Arc;
+use tokio::time::Duration;
+use tracing::{debug, warn};
+
+const HEARTBEAT_FUNCTION: &str = r#"
+CREATE OR REPLACE FUNCTION run_heartbeat()
+RETURNS VOID LANGUAGE plpgsql AS $$
+DECLARE
+    schema_exists BOOLEAN;
+    table_exists BOOLEAN;
+BEGIN
+    -- Check if schema exists
+    SELECT EXISTS(
+        SELECT schema_name
+        FROM information_schema.schemata
+        WHERE schema_name = 'tembo'
+    ) INTO schema_exists;
+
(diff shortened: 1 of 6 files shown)


tembo-operator/tests/integration_tests.rs
```diff
@@ -650,6 +650,17 @@ mod test {
...
+        // Check for heartbeat table and values
+        let sql_result = wait_until_psql_contains(
+            context.clone(),
+            coredb_resource.clone(),
+            "SELECT latest_heartbeat FROM tembo.heartbeat_table LIMIT 1".to_string(),
+            "postgres".to_string(),
+            true,
+        )
+        .await;
+        assert!(sql_result.success);
```
sjmiller609, line 662: nice test

tembo-operator/src/heartbeat.rs
```diff
@@ -0,0 +1,91 @@
...
+        SELECT table_name
+        FROM information_schema.tables
+        WHERE table_schema = 'tembo' AND table_name = 'heartbeat_table'
+    ) INTO table_exists;
+
+    -- Create table if it doesn't exist
+    IF NOT table_exists THEN
+        EXECUTE 'CREATE TABLE tembo.heartbeat_table (
+            latest_heartbeat TIMESTAMP NOT NULL
+        );';
```
sjmiller609, line 37: Please check on adding an index so that the table scan on WHERE last_heartbeat < is avoided. 
Check like this
```
EXPLAIN ANALYZE SELECT FROM tembo.heartbeat_table
        WHERE latest_heartbeat < (CURRENT_TIMESTAMP AT TIME ZONE ''UTC'' - INTERVAL ''7 days'');
```

tembo-operator/src/heartbeat.rs
```diff
@@ -0,0 +1,91 @@
...
+    END IF;
+
+    -- Insert current UTC timestamp into heartbeat_table
+    EXECUTE 'INSERT INTO tembo.heartbeat_table (latest_heartbeat)
+        VALUES (CURRENT_TIMESTAMP AT TIME ZONE ''UTC'');';
+
+    -- Delete entries older than 30 days
```
sjmiller609, line 44: match comment to query (7 v 30 days)

sjmiller609 (approved):
Nice! A couple of nits, then good to go 🎉

tembo-operator/src/heartbeat.rs
sjmiller609: basically I just mean check on the performance of this query if written every 60s times 7 days

nhudson: I was wondering if I needed to add one here.  I will get it added.  Thanks

tembo-operator/src/heartbeat.rs
nhudson: Ah thanks.  I originally had it 30 days, but that seemed excessive.

tembo-operator/src/heartbeat.rs
```diff
@@ -0,0 +1,94 @@
...
+
+    -- Create table if it doesn't exist
+    IF NOT table_exists THEN
+        EXECUTE 'CREATE TABLE tembo.heartbeat_table (
+            latest_heartbeat TIMESTAMP NOT NULL
+        );';
+    END IF;
+
+    -- Create index on latest_heartbeat column
+    EXECUTE 'CREATE INDEX idx_heartbeat ON tembo.heartbeat_table (latest_heartbeat);';
```
sjmiller609, line 41: does this need to be inside the if NOT block above?

sjmiller609 (approved):

tembo-operator/src/heartbeat.rs
nhudson: Yes it does.  Fixed now.  Thanks!

sjmiller609 (approved):

PR #370, opened 2023-11-04
fix: Instances stuck restarting for non-interruptible, long-running queries

sjmiller609:
If we have the restartedAt annotation, and the pod is older than this annotation, and the pod has been not ready for over 30 seconds, delete the pod.

This case is applicable to queries that do not handle the smart or fast postgres shutdown procedures. For example executing a PL/Python function that's long running is not interrupted even by fast shutdown, and instead needs immediate shutdown.


sjmiller609:
Working on a test for this

PR #386, opened 2023-11-16
Update operator storage configuration for App Services

EvanHStanton:
### 1. Update operator storage configuration for App Services
- (Cargo.toml) Bump version from 0.19.2 to 0.20.0.
- (manager.rs) Refactor generate_deployment function to accommodate app service storage configurations, including volume mounts.
- (types.rs) Introduce StorageConfig struct to handle volumes and volume mounts.
- (types.rs) Update AppService struct to include storage as an optional field.

### 2. Introduce Document Stack yaml for local testing 
- Add a new yaml/sample-document.yaml

### DIFF
--- a/charts/tembo-operator/templates/crd.yaml
+++ b/charts/tembo-operator/templates/crd.yaml
@@ -57,19 +57,6 @@ spec:
                         type: object
                       nullable: true
                       type: array
-                    ephemeral_volume_config:
-                      nullable: true
-                      properties:
-                        access_modes:
-                          items:
-                            type: string
-                          type: array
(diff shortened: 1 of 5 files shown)


sjmiller609 (approved):
Big change in the CRD itself, but I think this is a nice way to do it since it allows for more flexibility without changing the operator, like this would also work for persistent volumes for app services if we wanted those.

ianstanton (approved):
Looks good! Let's remember to  follow up with a test for this

ChuckHend (approved):

PR #394, opened 2023-11-22
Version lock linting

sjmiller609:


.github/workflows/cargo-test.yaml
```diff
@@ -64,10 +64,13 @@ jobs:
...
             ${{ matrix.path }}
-      - name: Cargo format
-        run: cd ${{ matrix.path }} && cargo +nightly fmt --all --check
-      - name: Clippy
-        run: cd ${{ matrix.path }} && cargo +stable clippy
+      - name: Cargo format and clippy
+        run: |
+          set -xe
+          cd ${{ matrix.path }}
+          cargo +nightly-2023-11-18 --version
```
sjmiller609, line 71: versions after this different lint on mac and linux for some reason? we can switch off hardcoding when that is fixed

